PR #6, opened 2018-10-25
Webots7.4.3

Giessen:
Make it work with pr2, ur10 and motoman.
*Fix some stl reading errors.
*Color will come from material instead.
*Texture images are processed and stored into <RobotName>_textures folder.
*Other minor changes.


fabienrohrer:
Duplicate with #4, wrong target (should by `master`).
I allow myself to close this PR.

PR #9, opened 2018-10-25
Move hard-coded collada directory to a requirement

### LINKED ISSUE #8: pycollada as a submodule or so
[`pycollada`](https://github.com/pycollada/pycollada) is currently copied in this repo which is a bad practice.
Including it as a submodule would be a better idea.

fabienrohrer:
Fix #8

`collada` directory is a hard-copy of the pycollada project.
This doesn't respect our [contributing rules](https://github.com/omichel/urdf2webots/blob/master/CONTRIBUTING.md#source-rules) ;-) 
Instead, it's much better to use the pip requirement file system as shown here.


fabienrohrer:
👍

PR #10, opened 2018-10-29
Add Travis tests

fabienrohrer:
Ditto

- [x] Launch automatically the tests from Travis
- [x] Add PEP8 test
- [ ] Add URDF exportation test


fabienrohrer:
A lot of ROS projects are now not using directly URDF, but XACRO which is basically an XML container with the possibility to write macros:
https://github.com/ros/xacro

This is the case for the RobotisOP3, the PR2 and the UR10, where the URDF should be extracted from XACRO to launch our exporter.

I tried to add XACRO as a submodule, and to only launch this part, but unfortunately, XACRO has dependencies on ROS :trollface:

fabienrohrer:
Obsolete

PR #12, opened 2019-03-07
Various Improvments

DavidMansolino:
Modifications:
  - [x] Various pep8 cleanup.
  - [x] Handle package relative path (e.g.: `<mesh filename="package://ur_description/meshes/ur5/visual/base.dae"/>`)
  - [x] Write mesh coordinates on one line.
  - [x] Switch to PBR.
  - [x] Fix missing shapes.
  - [x] Fix transformation issue.
  - [x] Add position sensors.
  - [x] Move math fucntion in math module
  - [x] Support basic sensors:
    - [x] Camera
    - [x] Lidar
    - [x] IMU (accelerometer+gyro+compass).
  - [x] PEP8 travis test.
  - [x] Improve README

PR #13, opened 2019-03-13
Export test

fabienrohrer:
- [x] Add a unit test to check some robots are exported as expected (replace #12)
- [x] Test `motoman`

---

- [x] Fix `--output` argument (which was not used ^^)
- [x] Fix return code


fabienrohrer:
For now, only the motoman arm export is tested.
Better than nothing.
The other modules would require to add xacro in travis which is not obvious.

👍

PR #14, opened 2019-03-26
Adding some checks

tahir1069:
The code expects for Indexed Faceset nodes a [triangular mesh conversion](https://www.cyberbotics.com/doc/reference/indexedfaceset). For .dae files or from blender these coordinates are given in a list, so I made this check to get the desired triangular mesh.

In joints the axis are needed for the calculation which are [optional in URDF](http://wiki.ros.org/urdf/XML/joint) so adding a small check also works.


omichel:
Thank you @tahir1069 for this contribution. However, you Python code didn't pass the CI test because it's not PEP8 compliant.

tahir1069:
ohh I saw that but was quite busy and had a plan for this week.
However so you did this thank you so much
Cheers!

PR #15, opened 2019-04-02
Support Normals Export

DavidMansolino:
Extend the exporter script to export normals as well.
Related to omichel/webots#312


DavidMansolino:
:+1:

PR #19, opened 2019-07-16
Optimize Mesh Import

DavidMansolino:
The goal of this PR is to optimize the big meshes importation and show the importation progress.


DavidMansolino:
:+1:

PR #28, opened 2019-09-24
Improve Modularization

DavidMansolino:
Improve the re-use of the module from another module.


DavidMansolino:
The creation and publishing of the package is very easy now:
```
python setup.py sdist
twine upload dist/* -r pypitest
```
Result in test server: https://test.pypi.org/project/urdf2webots/
Once this PR is approved I will publish to the production server of pypi (on a shared account).
The package is in 'restricted mode' (:wink:) it does not require any specific rights and is installed by default in the users directory.

For testing:

```
pip install -i https://test.pypi.org/simple/ urdf2webots
```
or in debug mode without cache:
```
pip3 install -v -v -v -i https://test.pypi.org/simple/ urdf2webots --upgrade --no-cache-dir
```

And then reffer to the readme: https://github.com/cyberbotics/urdf2webots/tree/improve-modularization#from-pip-1


:+1:

DavidMansolino:
:+1:

DavidMansolino:
:+1:

fabienrohrer:
👀

README.md
```diff
@@ -4,6 +4,14 @@ This tool converts URDF files into Webots PROTO files.
 
 ## Install
 
+### From pip
+
+```
+pip install urdf2webots
+```
+
```
fabienrohrer: On macOS, a pip binary cannot be run as-is, as explained here: https://stackoverflow.com/a/43368894/2210777

```suggestion

On macOS, export the pip binary path to the PATH: `export PATH="/Users/$USER/Library/Python/3.7/bin:$PATH"`

```

setup.py
```diff
@@ -0,0 +1,28 @@
...
+    license='Apache License, Version 2.0',
+    classifiers=[
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python',
+        'Topic :: Software Development'
+    ],
+    install_requires=[
+        "pycollada >= 0.6",
+        "Pillow"
```
fabienrohrer: Missing requirement:

```suggestion
        "Pillow",
        "numpy"
```

DavidMansolino:
This is caused by the space in the installation path of the Python. The shebang is automatically replaced by pip (https://github.com/cyberbotics/urdf2webots/pull/28/files#diff-32c8cb5ed2c36e878e1e703a17806a1eR1) at the installation but without adding quotes.
I will check if we have some control on this, but from what I saw we are not the only one to have this issue (and the only answer is 'edit the shebang after installation' :-/ ).

From the code (on Windows) it works fine, only the executable have problem.

DavidMansolino:
I think this is almost impossible to do, I have tried to put the path into quotes but python doesn't handle this.

And I found this:  https://lists.gnu.org/archive/html/bug-bash/2008-05/msg00052.html
> It's actually the kernel that interprets that line, not bash.  The
historical behavior is that space separates the interpreter from an
optional argument, and there is no escape mechanism, so there's no way
to specify an interpreter with a space in the path.  It's unlikely
that this would ever change, since that would break existing scripts
that rely on thecurrent behavior.

Furthermore, python is installed by default on Windows in the user AppData directory.  And for multi-user installation it is recommended to install it in C:\Python37 (https://python-forum.io/Thread-Part-1-Python-3-6-3-7-and-pip-installation-under-Windows?pid=51140#pid51140)

DavidMansolino:
I fixed the space issue on Windows, the new way to start a conversion directly is now:
```python -m urdf2webots.importer```
I suspect this will also fix the mac issue.

:+1:

omichel:
:eyes:

DavidMansolino:
Thank you for the review and test.
I will wait for the mac test too before merging.

DavidMansolino:
:+1:  (fixed conflicts with master)

Issue #29, opened 2019-09-25
Automatic Shape USE-DEF


DavidMansolino:
We should add a new argument to create automatically USE-DEF in case the same mesh is used several times.

DavidMansolino:
Fixed in #30

PR #30, opened 2019-09-26
Feature Automatic USE-DEF

DavidMansolino:
Implementation of #29


DavidMansolino:
A side effect of this improvement is that the mesh files are parsed only once improving greatly the conversion time (divided by two in the test for example).

DavidMansolino:
Geometries and appearances are now automatically USE-DEFed.

:+1:

PR #36, opened 2020-01-31
Add support for OBJ meshes

omichel:
This PR fixes the import of a URDF model created with [phobos](https://github.com/dfki-ric/phobos) and OBJ meshes.

In addition to OBJ support, this PR:
- Fixes the computation of the root link which was wrong.
- Fixes the import of the normal per vertex which was broken.
- Fixes a line length of python code that was exceeding our coding style limit (128 chars).
- Fixes a bug preventing the script to find the mesh files if not run from the URDF directory.
- Fixes the generation of DEF names which may contain spaces or dot characters (which are illegal for a VRML DEF name).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open('README.md', 'r') as fh:
 
 setuptools.setup(
     name='urdf2webots',
-    version='1.0.1',
+    version='1.0.2',
     author='Cyberbotics',
     author_email='support@cyberbotics.com',
     description='A converter between URDF and PROTO files.',
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open('README.md', 'r') as fh:
 
 setuptools.setup(
     name='urdf2webots',
-    version='1.0.2',
+    version='1.0.1',
     author='Cyberbotics',
     author_email='support@cyberbotics.com',
     description='A converter between URDF and PROTO files.',
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -250,17 +250,20 @@ class IMU():
         file.write(indentationLevel * indent + 'Accelerometer {\n')
         file.write(indentationLevel * indent + '  name "%s accelerometer"\n' % self.name)
         if self.gaussianNoise > 0:
-            file.write(indentationLevel * indent + '  lookupTable [-100 -100 %lf, 100 100 %lf]\n' % (-self.gaussianNoise / 100.0, self.gaussianNoise / 100.0))
+            file.write(indentationLevel * indent + '  lookupTable [-100 -100 %lf, 100 100 %lf]\n' %
+                       (-self.gaussianNoise / 100.0, self.gaussianNoise / 100.0))
         file.write(indentationLevel * indent + '}\n')
         file.write(indentationLevel * indent + 'Gyro {\n')
         file.write(indentationLevel * indent + '  name "%s gyro"\n' % self.name)
         if self.gaussianNoise > 0:
-            file.write(indentationLevel * indent + '  lookupTable [-100 -100 %lf, 100 100 %lf]\n' % (-self.gaussianNoise / 100.0, self.gaussianNoise / 100.0))
+            file.write(indentationLevel * indent + '  lookupTable [-100 -100 %lf, 100 100 %lf]\n' %
+                       (-self.gaussianNoise / 100.0, self.gaussianNoise / 100.0))
         file.write(indentationLevel * indent + '}\n')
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 
 setuptools.setup(
     name='urdf2webots',
-    version='1.0.1',
+    version='1.0.2',
```
DavidMansolino: This is not required as version 1.0.1 is not released yet.

urdf2webots/writeProto.py
```diff
@@ -224,6 +245,10 @@ def URDFBoundingObject(proto, link, level, boxCollision):
         proto.write(level * indent + '}\n')
 
 
+def defName(name):
+    return name.replace(' ', '_').replace('.', '_')
```
DavidMansolino: The name of this function lets think that it is a getter.
I would rather use an action in the name, e.g. `cleanDefName(name)`

urdf2webots/parserURDF.py
```diff
@@ -622,7 +675,8 @@ def getVisual(link, node):
             visual.geometry.sphere.radius = float(geometryElement.getElementsByTagName('sphere')[0].getAttribute('radius'))
             link.visual.append(visual)
         elif hasElement(geometryElement, 'mesh'):
-            meshfile = geometryElement.getElementsByTagName('mesh')[0].getAttribute('filename')
+            meshfile = os.path.normpath(os.path.join(path,
+                                                     geometryElement.getElementsByTagName('mesh')[0].getAttribute('filename')))
```
DavidMansolino: Are you sure that this will work if `filename` is an absolute path?

urdf2webots/parserURDF.py
```diff
@@ -413,6 +416,51 @@ def getSTLMesh(filename, node):
...
+                continue
+            elif header == 'v':  # vertex
+                (x, y, z) = body.split()
+                trimesh.coord.append([float(x), float(y), float(z)])
+            elif header == 'vn':  # normals
+                (x, y, z) = body.split()
+                trimesh.normal.append([float(x), float(y), float(z)])
+            elif header == 'vp':  # parameters, ignored
+                continue
+            elif header == 'vt':  # texture coordinate, ignored
```
DavidMansolino: Why do we ignore texture coordinate?

urdf2webots/parserURDF.py
omichel: I assumed it we never get an absolute path here. Am I wrong?

omichel: Anyhow I added a test in case we get an absolute path there, but I doubt this will happen.

DavidMansolino: From the documentation nothing prevents to use an absolute path (whereas `obj` file are not supposed to be supported):
> A trimesh element specified by a filename, and an optional scale that scales the mesh's axis-aligned-bounding-box. The recommended format for best texture and color support is Collada .dae files, though .stl files are also supported. The mesh file is not transferred between machines referencing the same model. It must be a local file. 

Furthermore, we might actually get absolute path due to:
https://github.com/cyberbotics/urdf2webots/blob/9b0e41069a8e73e9b518eb97c5f2d2d8deb65963/urdf2webots/importer.py#L73

Using `os.path.relpath` instead looks safer.

urdf2webots/parserURDF.py
omichel: I could not test it, so I preferred to not implement it yet. But I can implement it blindly...

omichel: (the OBJ format is a bit tricky about the number of parameters here, so the implementation is not straightforward)

DavidMansolino: Ok that's fine for me, I was just wondering if it was just a matter of adding the trimesh.texCoord (because from the documetnation of obj file it looked lije they were simply defined as `vt 1.0 0.0`).

omichel: Not that simple (from [wikipedia](https://en.wikipedia.org/wiki/Wavefront_.obj_file)):
```
# List of texture coordinates, in (u, [,v ,w]) coordinates, these will vary between 0 and 1. v, w are optional and default to 0.
  vt 0.500 1 [0]
```
But doable. I will try to implement it.

omichel:
Tested with this URDF: 
[gait2392_simbody.zip](https://github.com/cyberbotics/urdf2webots/files/4148338/gait2392_simbody.zip)

DavidMansolino:
Maybe we can add this one to the test suite.

DavidMansolino:
(because we are not going to test urdf with obj file very often)

omichel:
Good idea. I am doing this right now.

PR #39, opened 2020-02-03
Fix add test with obj meshes

omichel:
Added test with OBJ meshes.


DavidMansolino:
:eyes:

urdf2webots/parserURDF.py
```diff
@@ -457,7 +457,7 @@ def getOBJMesh(filename, node):
                 continue
             elif header == 'vt':  # texture coordinate
                 texCoord = body.split()
-                lenght = len(texCoord)
+                length = len(texCoord)
                 if length < 1:  # v argument is optional and defaults to 0
```
DavidMansolino: The length variable is used only once, it can be changed to:
```
if len(texCoord) < 1:  # v argument is optional and defaults to 0
```

urdf2webots/parserURDF.py
```diff
@@ -422,7 +422,7 @@ def getOBJMesh(filename, node):
     with open(filename, 'r') as file:
         trimesh = node.geometry.trimesh
         for line in file:
-            header, body = line.split(maxsplit=1)
+            header, body = line.split(' ', 1)
```
DavidMansolino: Just out of curiosity, what is the advantage of this change?

omichel: Compatibility with Python 2.7

PR #43, opened 2020-05-06
Fix Empty Def Name

DavidMansolino:
This PR avoids using empty def names.


DavidMansolino:
@tahir1069 may I ask you to try this PR?

tahir1069:
@DavidMansolino this fixes the problem the proto works now. Thank you so much.

DavidMansolino:
Thank you for the feedback 😃

PR #53, opened 2020-08-01
add feature --multi-file

Vidicon:
Adds the feature to export the meshes as separate proto files. 
I tested it on the [turtlebot3_description](http://wiki.ros.org/turtlebot3_description) package and seems to be working as intended.

Most likely needs some more (cleanup) work before merging.


omichel:
Although this PR is perfectly fine, it could have been easier and handy to export the meshes using the new [Mesh](https://cyberbotics.com/doc/reference/mesh) node of Webots R2020b so that the same mesh files can be used by both the Webots PROTO and the URDF file.

PR #70, opened 2020-08-13
Make multi-file mesh names unique

Simon-Steinmann:
Adds the main proto file name in front of the mesh names


urdf2webots/parserURDF.py
```diff
@@ -850,7 +850,9 @@ def getLink(node, path):
         getVisual(link, node, path)
     if hasElement(node, 'collision'):
         getCollision(link, node, path)
-    return link
+    if not any([hasElement(node, 'inertial'), hasElement(node, 'visual'), hasElement(node, 'inecollisionrtial')]):
```
DavidMansolino, line 853:
```suggestion
    if not any([hasElement(node, 'inertial'), hasElement(node, 'visual'), hasElement(node, 'collision')]):
```

Simon-Steinmann:
Did you remove the effort limit fix on purpose?

DavidMansolino:
> Did you remove the effort limit fix on purpose?

Yes, I found a simpler solution just by changing the default value here: https://github.com/cyberbotics/urdf2webots/pull/70/files#diff-725bae3adc35161533858377f44cdbfaR196

Let me know if this cause some issue in your test cases.

Simon-Steinmann:
Did you do a different implementation of the torque fix, or did it just get lost somewhere?

Simon-Steinmann:
ahh gotcha

DavidMansolino:
> Did you do a different implementation of the torque fix, or did it just get lost somewhere?

Yes just changing the default value here https://github.com/cyberbotics/urdf2webots/pull/70/files#diff-725bae3adc35161533858377f44cdbfaR196 fixes the issue since if the `effort` is not defined it will use this default value.

PR #71, opened 2020-08-13
Add rotation Argument

Simon-Steinmann:
Adds the option to parse the rotation of the PROTO file

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -28,4 +28,4 @@ optParser.add_option('--rotation', dest='itiRotation', default='0 1 0 0',
 options, args = optParser.parse_args()
 
 convert2urdf(options.inFile, options.outFile, options.normal, options.boxCollision, options.disableMeshOptimization,
-                options.enableMultiFile, options.staticBase, options.toolSlot, options.itiRotation)
+             options.enableMultiFile, options.staticBase, options.toolSlot, options.itiRotation)
--- a/demo.py
+++ b/demo.py
@@ -23,9 +23,9 @@ optParser.add_option('--static-base', dest='staticBase', action='store_true', de
                      help='If set, the base link will have the option to be static (disable physics)')
 optParser.add_option('--tool-slot', dest='toolSlot', default=None,
                      help='Specify the link that you want to add a tool slot to (exact link name from urdf)')
-optParser.add_option('--rotation', dest='itiRotation', default='0 1 0 0',
+optParser.add_option('--rotation', dest='initRotation', default='0 1 0 0',
                      help='Set the rotation field of your PROTO file.)')
 options, args = optParser.parse_args()
 
 convert2urdf(options.inFile, options.outFile, options.normal, options.boxCollision, options.disableMeshOptimization,
-             options.enableMultiFile, options.staticBase, options.toolSlot, options.itiRotation)
+             options.enableMultiFile, options.staticBase, options.toolSlot, options.initRotation)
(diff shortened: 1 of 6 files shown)


demo.py
```diff
@@ -23,7 +23,9 @@
                      help='If set, the base link will have the option to be static (disable physics)')
 optParser.add_option('--tool-slot', dest='toolSlot', default=None,
                      help='Specify the link that you want to add a tool slot to (exact link name from urdf)')
+optParser.add_option('--rotation', dest='itiRotation', default='0 1 0 0',
+                        help='Set the rotation field of your PROTO file.)')
 options, args = optParser.parse_args()
 
 convert2urdf(options.inFile, options.outFile, options.normal, options.boxCollision, options.disableMeshOptimization,
-             options.enableMultiFile, options.staticBase, options.toolSlot)
+                options.enableMultiFile, options.staticBase, options.toolSlot, options.itiRotation)
```
DavidMansolino, line 31:
```suggestion
             options.enableMultiFile, options.staticBase, options.toolSlot, options.itiRotation)
```

demo.py
```diff
@@ -23,7 +23,9 @@
                      help='If set, the base link will have the option to be static (disable physics)')
 optParser.add_option('--tool-slot', dest='toolSlot', default=None,
                      help='Specify the link that you want to add a tool slot to (exact link name from urdf)')
+optParser.add_option('--rotation', dest='itiRotation', default='0 1 0 0',
+                        help='Set the rotation field of your PROTO file.)')
```
DavidMansolino, line 27:
```suggestion
                     help='Set the rotation field of your PROTO file.)')
```

urdf2webots/importer.py
```diff
@@ -47,14 +47,17 @@ def mkdirSafe(directory):
 
 
 def convert2urdf(inFile, outFile=None, normal=False, boxCollision=False,
-                 disableMeshOptimization=False, enableMultiFile=False, staticBase=False, toolSlot=None):
+                 disableMeshOptimization=False, enableMultiFile=False, staticBase=False, toolSlot=None, initRotation='0 1 0 0'):
     if not os.path.exists(inFile):
         sys.exit('Input file "%s" does not exists.' % inFile)
+    if not type(initRotation)==str or len(initRotation.split())!=4:
```
DavidMansolino, line 53:
```suggestion
    if not type(initRotation) == str or len(initRotation.split()) != 4:
```

urdf2webots/writeProto.py
```diff
@@ -56,7 +57,7 @@ def declaration(proto, robotName):
     """Prototype declaration."""
     proto.write('PROTO ' + robotName + ' [\n')
     proto.write('  field  SFVec3f     translation     0 0 0\n')
-    proto.write('  field  SFRotation  rotation        0 1 0 0\n')
+    proto.write('  field  SFRotation  rotation        ' + initRotation + '\n')
```
DavidMansolino, line 60: I would rather pass this as an argument of the `declaration` function instead of a local variable.

Simon-Steinmann: sure, go ahead. This was the quickest solution I came up with. A cleaner one is of course better

Simon-Steinmann: I can do it too if you want

DavidMansolino: As you prefeer, I can do it but tomorrow morning,

Simon-Steinmann: done

omichel:
The tutorial should probably be updated as well. The list of arguments has one more entry and the rotation step is not needed any more 😉.

DavidMansolino:
Yes, and it should probably be avoided to duplicate the argument list both in the tutorial and README...

PR #73, opened 2020-08-20
name fix/add, material fix

Simon-Steinmann:
This PR includes several fixes and changes:
- name field, which specifies the name of the base node. default gets set to the robot name
- fixed a bug, where materials were not parsed, when they had the name="" (mirobot convertion as example)
- fixed the toolSlot generation. When adding a gripper, it fell off. added a physics and bounding object node
- specifying a output file, broke the conversion, as the filename and proto name has to be identical. I split it up into outputPath and name, The name get's take for both, the file AND the protos definition

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -7,25 +7,27 @@ import optparse
 from urdf2webots.importer import convert2urdf
 
 
-optParser = optparse.OptionParser(usage='usage: %prog --input=my_robot.urdf [options]')
-optParser.add_option('--input', dest='inFile', default='', help='Specifies the urdf file to convert.')
-optParser.add_option('--output', dest='outFile', default='', help='Specifies the name of the resulting PROTO file.')
-optParser.add_option('--normal', dest='normal', action='store_true', default=False,
-                     help='If set, the normals are exported if present in the URDF definition.')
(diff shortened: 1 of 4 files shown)


Simon-Steinmann:
i have not added the readme yet, as I'm sure there will be suggestions and changes ;)

demo.py
```diff
@@ -7,25 +7,27 @@
...
-                     help='If set, the base link will have the option to be static (disable physics)')
-optParser.add_option('--tool-slot', dest='toolSlot', default=None,
-                     help='Specify the link that you want to add a tool slot to (exact link name from urdf)')
-optParser.add_option('--rotation', dest='initRotation', default='0 1 0 0',
-                     help='Set the rotation field of your PROTO file.)')
-options, args = optParser.parse_args()
+if __name__ == '__main__':
+    optParser = optparse.OptionParser(usage='usage: %prog --input=my_robot.urdf [options]')
+    optParser.add_option('--input', dest='inFile', default='', help='Specifies the urdf file to convert.')
+    optParser.add_option('--nane', dest='robotName', default=None, help='The exact name of the webots model and filename.')
```
DavidMansolino, line 13:
```suggestion
    optParser.add_option('--name', dest='robotName', default=None, help='The exact name of the webots model and filename.')
```

PR #74, opened 2020-08-21
Automatic conversion tool + urdf library w/ meshes

Simon-Steinmann:
Adds a automatic batch-conversion tool. How it works:

in the root directory, run:
`python batch_conversion.py `

This will take any URDF file in the `automatic-conversion/urdf `folder and convert it into a proto file and save it in a unique folder in the `automatic-conversion/protos ` directory. It will have the same nested folder structure as in the `automatic-conversion/urdf ` folder. The options for urdf conversion are defined in a .json file, which gets automatically created for any new urdf file added.

Once done converting, it will delete everything in the `automatic-conversion/ExtraProjectTest ` folder, and add the newly converted files there. This directory can be added to Webots' `Extra Project Path`. This way, conversions can be easily tested, without manually copying them somewhere. Simply requires Webots restart.

### DIFF
--- a/batch_conversion.py
+++ b/batch_conversion.py
@@ -1,6 +1,37 @@
 #!/usr/bin/env python3
 
-"""URDF files to Webots PROTO converter."""
+"""URDF files to Webots PROTO batch-converter.
+------------------------------------------------------------------------
+This script searches for all urdf files in the <input> directory, converts
+them to proto files and saves them in the output directory. It retains
+the folder structure of the input directory while doing so.
+
+The first time a urdf file gets converted, this script creates a config
+.json file next to it. Here you can specify any arguments you want
+to parse to that specific file conversion. The definiton of the default
+config is in the code below and needs to be updated when features get added.
+
+For each batch-conversion, a new (unique) folder is generated inside the
+output folder, where the conversion is being placed in. This way, previous
+conversions are not overridden. In addition, the converted files override
+those in '<output>/ExtraProjectTest'  and the TestAllRobots.wbt world is
+created and placed there, containing all converted models, placed in a grid.
+
+To test the conversions, open webots and change your
+Tools>Prefernces>Extra_projects_path to '<output>/ExtraProjectTest' and
+open the generated TestAllRobots.wbt. From now on, after another conversion,
+you simply have to reload the world.
+
+Example:
+<input>/robots/kuka/protos/kr5.urdf
+gets exportet to
+'<output>/<unique_folder>/robots/kuka/protos/kr5.proto'
+and
+'<output>/ExtraProjectTest/robots/kuka/protos/kr5.proto'
+
+Note: Webots will only automatically find your proto file, if it is inside a
+'protos' folder.
+"""
 
 import optparse
 import datetime
@@ -187,6 +218,9 @@ class BatchConversion():
                 print(urdf)
 
     def create_test_world(self, spacing=3):
+        '''Create a TestWorld (<output>/ExtraProjectTest/TestAllRobots.wbt')
+        where every converted robot or model is placed in a grid, with the
+        spacing representing the grid size'''
         n_models = len(self.urdf_files)
         n_row = math.ceil(n_models ** 0.5)  # round up the sqrt of the number of models
         grid_size = spacing * (n_row + 1)


DavidMansolino:
That looks very nice and convenient, thank you for sharing this!

Note that the 'batch_conversion.py' file does not respect PEP8 (which is causing the automatic tests to fail): https://travis-ci.com/github/cyberbotics/urdf2webots/jobs/376314699#L303

Simon-Steinmann:
got it, didn't know that was a rule. Will keep in mind.

Simon-Steinmann:
Test the tool by following these steps:
1. checkout to this branch an pull
2. run `python batch_conversion.py `
3. Once completed, open Webots and change the Extra projects path to` urdf2webots/automatic_conversion/ExtraProjectTest` (of course the full path in your system)
4. Open the testworld located at ` urdf2webots/automatic_conversion/ExtraProjectTest/TestAllRobots.wbt`
This will take a moment to load, as it loads every converted model into a dynamically created world.

DavidMansolino (commented):
We discussed with the @cyberbotics/maintainers team and we believe it would be cleaner to put the content of the `automatic_conversion` folder in a root `source` folder in the https://github.com/cyberbotics/community-projects repository, this way, the source and result remain close.
This will also keep this repository clean and light (which is important as this repository is used in other projects as a submodule).

Simon-Steinmann:
That is a good point. I agree. Did you test it though? have you feedback on the code and process itself?

DavidMansolino:
I didn't test it yet, but I had a look at the code and it looks like a very nice and useful tool!

urdf2webots/writeProto.py
```diff
@@ -155,18 +155,39 @@ def URDFLink(proto, link, level, parentList, childList, linkList, jointList, sen
             proto.write((level + 1) * indent + 'physics Physics {\n')
             proto.write((level + 2) * indent + 'density -1\n')
             proto.write((level + 2) * indent + 'mass %lf\n' % link.inertia.mass)
-            if link.inertia.ixx > 0.0 and link.inertia.iyy > 0.0 and link.inertia.izz > 0.0:
-                proto.write((level + 2) * indent + 'centerOfMass [ %lf %lf %lf ]\n' % (link.inertia.position[0],
+            #  if link.inertia.position != [0.0, 0.0, 0.0]:
```
Simon-Steinmann, line 158: This check causes issues. A link with mass can have no origin-translation for the inertia. We already check if the link has mass. That is enough.

urdf2webots/writeProto.py
```diff
@@ -473,8 +503,8 @@ def URDFShape(proto, link, level, normal=False):
             if name is None:
                 if visualNode.geometry.name is not None:
                     name = computeDefName(visualNode.geometry.name)
-            if visualNode.geometry.defName is None:
-                name = robotNameMain + '_' + name if robotNameMain else name
+            name = robotNameMain + '_' + name if robotNameMain else name
```
Simon-Steinmann, line 506: We have to define the name before the condition, otherwise the adjusted path is not used, causing errors

Simon-Steinmann:
The two comments I left are the only changes specific to this, the rest is updates from recent changes (pulled the newest master just before and added these 2 changes)

I also added parsable input and output paths to the batch_conversion.py. You should call it now like this:
`python batch_conversion.py --input=/home/simon/community-projects/sources --output=/home/simon/community-projects/converted `

I pushed the urdf sources I'm currently using. You can pull them here:
https://github.com/cyberbotics/community-projects/tree/Simon-Steinmann-sources

urdf2webots/writeProto.py
DavidMansolino: I am not sure to understand your point, we are using later `link.inertia.position`, so if it doesn't exist it will cause issues too, no?

DavidMansolino:
Can you please update with master and resolve conflicts?

urdf2webots/writeProto.py
Simon-Steinmann: we dont check whether it exists, we check whether it is not [0, 0, 0]
Many urdf files have at least one link, that has a [0, 0, 0] translation for its inertia. In this case, we don't write the center of mass and Webots complains. As soon as we have a defined mass, we have to have a center of mass. We are already checking if we have mass, so this must be included, without additional checks.

Simon-Steinmann:
I updated with master

Simon-Steinmann:
The two comments I left are the only changes specific to this, the rest is updates from recent changes (pulled the newest master just before and added these 2 changes)

I also added parsable input and output paths to the batch_conversion.py. You should call it now like this:
`python batch_conversion.py --input=/home/simon/community-projects/sources --output=/home/simon/community-projects/converted `

I pushed the urdf sources I'm currently using. You can pull them here:
https://github.com/cyberbotics/community-projects/tree/Simon-Steinmann-sources

DavidMansolino:
> I updated with master, but I will fix the pep8 errors.

But from the diff it looks like it is not the case, because it contains changes from other PR (e.g. these changes https://github.com/cyberbotics/urdf2webots/pull/74/files#diff-68c204ae057e777029c1e7dbd77d1f03R393-R402 are from #76).

Simon-Steinmann:
These changes are compared to what this branch started out with. What is the proper way to do this? Using the updated master as the base?

DavidMansolino:
Yes, you should update your local master branch, then merge it with this branch and finally push this branch, here is how I do this (but that's not the only way):
```
git checkout master
git pull
git checkout Simon-Steinmann-batch-conversion
git merge master
git push origin Simon-Steinmann-batch-conversion
```

Simon-Steinmann:
Is there a way to see the difference in the expected and generated file?

DavidMansolino:
Sure, the simplest solution is to run the tests locally on your computer:
```
pip install -r requirements.txt
pip install -r tests/requirements.txt
python -m unittest discover -s tests
```
Then you can compare the files in: https://github.com/cyberbotics/urdf2webots/tree/master/tests/expected
With the files generated in `tests/results`.

Simon-Steinmann:
Ahh I had to 
git submodule update --init --recursive
to run the test.

The files are exactly the same, except for the line (in each file:)
`# Extracted from: /home/david/urdf2webots/tests/sources/motoman/motoman_sia20d_support/urdf/sia20d.urdf`

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        self.convert_all_urdfs()
+        self.print_end_report()
+
+    def create_proto_dir(self):
+        # create a new unique directory for our conversions
+        print(self.protoTargetDir)
+        os.makedirs(self.protoTargetDir)
+
+    def replace_ExtraProjectPath(self):
+        # this copies the converted files and puts them in the "automatic_conversion/ExtraProjectTest" directory
```
DavidMansolino, line 88:
```suggestion
        # this copies the converted files and puts them in the "<protoTargetDir>/ExtraProjectTest" directory
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        os.chdir(self.urdf2webots_path)
+
+
+if __name__ == '__main__':
+    optParser = optparse.OptionParser(usage='usage: %prog  [options]')
+    optParser.add_option('--input', dest='sourcePath', default='', help='Specifies the urdf file to convert.')
+    optParser.add_option('--output', dest='outPath', default='', help='Specifies the path and, if ending in ".proto", name of the resulting PROTO file.'
+                         ' The filename minus the .proto extension will be the robot name.')
+    optParser.add_option('--force-mesh-optimization', dest='forceMesh', action='store_true', default=False, help='Set if mesh-optimization should be turned on for all conversions. This will take much longer!')
+    optParser.add_option('--no-project-override', dest='extraProj', action='store_true', default=False, help='Set if new conversions should NOT replace existing ones in "automatic_conversion/ExtraProjectTest".')
```
DavidMansolino, line 242:
```suggestion
    optParser.add_option('--no-project-override', dest='extraProj', action='store_true', default=False, help='Set if new conversions should NOT replace existing ones in "<input>/ExtraProjectTest".')
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
+#!/usr/bin/env python
+
+"""URDF files to Webots PROTO converter."""
```
DavidMansolino, line 3: It would be nice to give a small explanation here about how to use this script.

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        worldFile.write('TexturedBackground {\n')
+        worldFile.write('}\n')
+        worldFile.write('TexturedBackgroundLight {\n')
+        worldFile.write('}\n')
+        worldFile.write('RectangleArena {\n')
+        worldFile.write('  floorSize ' + str(grid_size) + ' ' + str(grid_size) + '\n')
+        worldFile.write('  floorTileSize 0.25 0.25\n')
+        worldFile.write('  wallHeight 0.05\n')
+        worldFile.write('}\n')
+        filenames = []  # List which will store all of the full filepaths.
```
DavidMansolino, line 214:
```suggestion
```

Never used ;-)

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        # Report dict we can print at the end
+        self.EndReportMessage = {
+            'updateConfigs': [],
+            'newConfigs': [],
+            'converted': [],
+            'failed': []
+        }
+
+    def update_and_convert(self):
+        # Make sure all configs exist and are up to date, then convert all URDF files
```
DavidMansolino, line 77:
```suggestion
        """Make sure all configs exist and are up to date, then convert all URDF files."""
```

Python function description should looks like this.

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        self.check_all_configs()
+        self.convert_all_urdfs()
+        self.print_end_report()
+
+    def create_proto_dir(self):
+        # create a new unique directory for our conversions
```
DavidMansolino, line 83:
```suggestion
        """Create a new unique directory for our conversions."""
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        # every time
+        destination = os.path.dirname(self.protoTargetDir) + '/ExtraProjectTest'
+        if os.path.isfile(destination) or os.path.islink(destination):
+            os.remove(destination)  # remove the file
+        elif os.path.isdir(destination):
+            shutil.rmtree(destination)  # remove dir and all contains
+        shutil.copytree(self.protoTargetDir,  destination)
+
+    def update_config(self, configFile, config=None):
+        # makes sure the existing configs are in the same format as the default. Existing options are conserved
```
DavidMansolino, line 99:
```suggestion
        """Makes sure the existing configs are in the same format as the default, existing options are conserved."""
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        for key in new_config.keys():
+            try:
+                new_config[key] = config[key]
+            except:
+                pass
+        with open(configFile, 'w') as outfile:
+            json.dump(new_config, outfile, indent=4, sort_keys=True)
+
+    def check_all_configs(self):
+        # makes sure ever URDF file has a config and it is up do date
```
DavidMansolino, line 110:
```suggestion
        """Makes sure ever URDF file has a config and it is up to date."""
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+                    update_config(configFile, config)
+                else:
+                    print('Config up to date - ', configFile)
+            except:
+                print('Generating new config for - ' + os.path.splitext(self.urdf_files[i])[0])
+                self.EndReportMessage['newConfigs'].append(configFile)
+                update_config(configFile)
+
+    def convert_all_urdfs(self):
+        # convertes all URDF files according to their .json config files
```
DavidMansolino, line 130:
```suggestion
        """Converts all URDF files according to their '.json' config files."""
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+                    toolSlot=config['toolSlot'],
+                    initRotation=config['initRotation'])
+                self.EndReportMessage['converted'].append(self.urdf_files[i])
+            except Exception as e:
+                print(e)
+                self.EndReportMessage['failed'].append(self.urdf_files[i])
+            print('---------------------------------------')
+
+    def print_end_report(self):
+        # print the Report
```
DavidMansolino, line 164:
```suggestion
        """Print the report."""
```

urdf2webots/writeProto.py
DavidMansolino: This adds useless center of mass new lines in the test (which is probably the cause of the failure):
```
centerOfMass [ 0.000000 0.000000 0.000000 ]
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+    'boxCollision': True,
+    'disableMeshOptimization': True,
+    'enableMultiFile': True,
+    'staticBase': True,
+    'toolSlot': None,
+    'initRotation': "1 0 0 -1.5708"
+}
+
+
+class batchConversion():
```
DavidMansolino, line 39:
```suggestion
class BatchConversion():
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+    optParser = optparse.OptionParser(usage='usage: %prog  [options]')
+    optParser.add_option('--input', dest='sourcePath', default='', help='Specifies the urdf file to convert.')
+    optParser.add_option('--output', dest='outPath', default='', help='Specifies the path and, if ending in ".proto", name of the resulting PROTO file.'
+                         ' The filename minus the .proto extension will be the robot name.')
+    optParser.add_option('--force-mesh-optimization', dest='forceMesh', action='store_true', default=False, help='Set if mesh-optimization should be turned on for all conversions. This will take much longer!')
+    optParser.add_option('--no-project-override', dest='extraProj', action='store_true', default=False, help='Set if new conversions should NOT replace existing ones in "automatic_conversion/ExtraProjectTest".')
+    optParser.add_option('--update-cfg', dest='cfgUpdate', action='store_true', default=False, help='No conversion. Only updates or creates .json config for every URDF file in "automatic_conversion/urdf".')
+
+    options, args = optParser.parse_args()
+    bc = batchConversion(options.sourcePath, options.outPath)
```
DavidMansolino, line 245:
```suggestion
    bc = BatchConversion(options.sourcePath, options.outPath)
```

batch_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+        for urdf in self.EndReportMessage['converted']:
+            print(urdf)
+        if len(self.EndReportMessage['failed']) > 0:
+            print('---------------------------------------')
+            print('Failedd URDF conversions:')
+            print('---------------------------------------')
+            for urdf in self.EndReportMessage['failed']:
+                print(urdf)
+
+    def create_test_world(self, spacing=3):
```
DavidMansolino, line 189: A small description of what this function actually does would be welcome.

batch_conversion.py
Simon-Steinmann: noted, I'll work on that

batch_conversion.py
Simon-Steinmann: relict from earlier version ^^

Simon-Steinmann:
Did the test work for you? With conversion and loading the TestWorld.wbt?

DavidMansolino (commented):
> Ahh I had to
> git submodule update --init --recursive
> to run the test.
> 
> The files are exactly the same, except for the line (in each file:)
> `# Extracted from: /home/david/urdf2webots/tests/sources/motoman/motoman_sia20d_support/urdf/sia20d.urdf`

When I run it, I have the following diff:
```diff
diff --git a/tests/expected/Human.proto b/tests/expected/Human.proto
index b0fa3da..0564c41 100644
--- a/tests/expected/Human.proto
+++ b/tests/expected/Human.proto
@@ -2,7 +2,7 @@
 # license: Apache License 2.0
 # license url: http://www.apache.org/licenses/LICENSE-2.0
 # This is a proto file for Webots for the Human
-# Extracted from: /home/travis/urdf2webots/tests/sources/gait2392_simbody/urdf/human.urdf
+# Extracted from: /home/david/urdf2webots/tests/sources/gait2392_simbody/urdf/human.urdf
 
 PROTO Human [
   field  SFVec3f     translation     0 0 0
@@ -276,6 +276,7 @@ PROTO Human [
                                   physics Physics {
                                     density -1
                                     mass 0.216600
+                                    centerOfMass [ 0.000000 0.000000 0.000000 ]
                                     inertiaMatrix [
                                       0.000100 0.000200 0.001000
                                       0.000000 0.000000 0.000000
@@ -298,6 +299,7 @@ PROTO Human [
                             physics Physics {
                               density -1
                               mass 1.250000
+                              centerOfMass [ 0.000000 0.000000 0.000000 ]
                               inertiaMatrix [
                                 0.001400 0.003900 0.004100
                                 0.000000 0.000000 0.000000
@@ -320,6 +322,7 @@ PROTO Human [
                       physics Physics {
                         density -1
                         mass 0.100000
+                        centerOfMass [ 0.000000 0.000000 0.000000 ]
                         inertiaMatrix [
                           0.001000 0.001000 0.001000
                           0.000000 0.000000 0.000000
@@ -342,6 +345,7 @@ PROTO Human [
                 physics Physics {
                   density -1
                   mass 3.707500
+                  centerOfMass [ 0.000000 0.000000 0.000000 ]
                   inertiaMatrix [
                     0.050400 0.005100 0.051100
                     0.000000 0.000000 0.000000
@@ -364,6 +368,7 @@ PROTO Human [
           physics Physics {
             density -1
             mass 9.301400
+            centerOfMass [ 0.000000 0.000000 0.000000 ]
             inertiaMatrix [
               0.133900 0.035100 0.141200
               0.000000 0.000000 0.000000
@@ -601,6 +606,7 @@ PROTO Human [
                                   physics Physics {
                                     density -1
                                     mass 0.216600
+                                    centerOfMass [ 0.000000 0.000000 0.000000 ]
                                     inertiaMatrix [
                                       0.000100 0.000200 0.001000
                                       0.000000 0.000000 0.000000
@@ -623,6 +629,7 @@ PROTO Human [
                             physics Physics {
                               density -1
                               mass 1.250000
+                              centerOfMass [ 0.000000 0.000000 0.000000 ]
                               inertiaMatrix [
                                 0.001400 0.003900 0.004100
                                 0.000000 0.000000 0.000000
@@ -645,6 +652,7 @@ PROTO Human [
                       physics Physics {
                         density -1
                         mass 0.100000
+                        centerOfMass [ 0.000000 0.000000 0.000000 ]
                         inertiaMatrix [
                           0.001000 0.001000 0.001000
                           0.000000 0.000000 0.000000
@@ -667,6 +675,7 @@ PROTO Human [
                 physics Physics {
                   density -1
                   mass 3.707500
+                  centerOfMass [ 0.000000 0.000000 0.000000 ]
                   inertiaMatrix [
                     0.050400 0.005100 0.051100
                     0.000000 0.000000 0.000000
@@ -689,6 +698,7 @@ PROTO Human [
           physics Physics {
             density -1
             mass 9.301400
+            centerOfMass [ 0.000000 0.000000 0.000000 ]
             inertiaMatrix [
               0.133900 0.035100 0.141200
               0.000000 0.000000 0.000000
@@ -926,6 +936,7 @@ PROTO Human [
                                   physics Physics {
                                     density -1
                                     mass 0.457500
+                                    centerOfMass [ 0.000000 0.000000 0.000000 ]
                                     inertiaMatrix [
                                       0.000892 0.000547 0.001340
                                       0.000000 0.000000 0.000000
@@ -948,6 +959,7 @@ PROTO Human [
                             physics Physics {
                               density -1
                               mass 0.607500
+                              centerOfMass [ 0.000000 0.000000 0.000000 ]
                               inertiaMatrix [
                                 0.002962 0.000618 0.003213
                                 0.000000 0.000000 0.000000
@@ -970,6 +982,7 @@ PROTO Human [
                       physics Physics {
                         density -1
                         mass 0.607500
+                        centerOfMass [ 0.000000 0.000000 0.000000 ]
                         inertiaMatrix [
                           0.002962 0.000618 0.003213
                           0.000000 0.000000 0.000000
@@ -992,6 +1005,7 @@ PROTO Human [
                 physics Physics {
                   density -1
                   mass 2.032500
+                  centerOfMass [ 0.000000 0.000000 0.000000 ]
                   inertiaMatrix [
                     0.011946 0.004121 0.013409
                     0.000000 0.000000 0.000000
@@ -1186,6 +1200,7 @@ PROTO Human [
                                   physics Physics {
                                     density -1
                                     mass 0.457500
+                                    centerOfMass [ 0.000000 0.000000 0.000000 ]
                                     inertiaMatrix [
                                       0.000892 0.000547 0.001340
                                       0.000000 0.000000 0.000000
@@ -1208,6 +1223,7 @@ PROTO Human [
                             physics Physics {
                               density -1
                               mass 0.607500
+                              centerOfMass [ 0.000000 0.000000 0.000000 ]
                               inertiaMatrix [
                                 0.002962 0.000618 0.003213
                                 0.000000 0.000000 0.000000
@@ -1230,6 +1246,7 @@ PROTO Human [
                       physics Physics {
                         density -1
                         mass 0.607500
+                        centerOfMass [ 0.000000 0.000000 0.000000 ]
                         inertiaMatrix [
                           0.002962 0.000618 0.003213
                           0.000000 0.000000 0.000000
@@ -1252,6 +1269,7 @@ PROTO Human [
                 physics Physics {
                   density -1
                   mass 2.032500
+                  centerOfMass [ 0.000000 0.000000 0.000000 ]
                   inertiaMatrix [
                     0.011946 0.004121 0.013409
                     0.000000 0.000000 0.000000
@@ -1274,6 +1292,7 @@ PROTO Human [
           physics Physics {
             density -1
             mass 26.826600
+            centerOfMass [ 0.000000 0.000000 0.000000 ]
             inertiaMatrix [
               1.474500 0.755500 1.431400
               0.000000 0.000000 0.000000
@@ -1296,6 +1315,7 @@ PROTO Human [
     physics Physics {
       density -1
       mass 11.777000
+      centerOfMass [ 0.000000 0.000000 0.000000 ]
       inertiaMatrix [
         0.102800 0.087100 0.057900
         0.000000 0.000000 0.000000
```

batch_conversion.py
Simon-Steinmann: What logic are you following when using """ instead of #?

DavidMansolino: This one: https://www.python.org/dev/peps/pep-0257/

batch_conversion.py
Simon-Steinmann: will do

Simon-Steinmann:
Can you handle the test files? For me only the extracted from was different. But it has to match the setup of the automated test by travis right?

DavidMansolino:
> Can you handle the test files? For me only the extracted from was different. But it has to match the setup of the automated test by travis right?

Just did: https://github.com/cyberbotics/urdf2webots/pull/74/commits/0c9e3e064a2bf20d23c2b783cff6a769c0fc51ed

batch_conversion.py
```diff
@@ -0,0 +1,253 @@
...
+        print('---------------------------------------')
+        for i in range(len(self.urdf_files)):
+            configFile = os.path.splitext(self.urdf_files[i])[0] + '.json'
+            try:
+                with open(configFile) as json_file:
+                    config = json.load(json_file)
+                if config.keys() != default_config.keys():
+                    print('Updating config (old settings will be carried over) - ', configFile)
+                    self.EndReportMessage['updateConfigs'].append(configFile)
+                    update_config(configFile, config)
```
DavidMansolino, line 121:
```suggestion
                    self.update_config(configFile, config)
```

batch_conversion.py
```diff
@@ -0,0 +1,253 @@
...
+                else:
+                    print('Config up to date - ', configFile)
+            except:
+                print('Generating new config for - ' + os.path.splitext(self.urdf_files[i])[0])
+                self.EndReportMessage['newConfigs'].append(configFile)
+                update_config(configFile)
```
DavidMansolino, line 127:
```suggestion
                self.update_config(configFile)
```

batch_conversion.py
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env python
```
DavidMansolino, line 1:
```suggestion
#!/usr/bin/env python3
```

Seems not to be compatible with python2 (this is not a problem, we don't really care of python2 anymore).

urdf2webots/writeProto.py
Simon-Steinmann: We have to think about this critically!

Simon-Steinmann:
Added the descriptions.

DavidMansolino:
image

It seems the robot are added twice to the test world, do you have the same behavior?

Simon-Steinmann:
no, only once for me

PR #76, opened 2020-08-21
ColladaMesh generation fix for LineSet

Simon-Steinmann:
Importing a LineSet causes an empty mesh to get created, throwing an error on import. My fix skips the import of Linesets and only imports PolyList and TriangleSet


DavidMansolino:
You should make sure when you create the new branch that you create it from the 'master' branch (and that you didn't commit anything to the master branch directly, to be sure the simplest is to delete your local master branch and re-fetch the one of the repository).

Simon-Steinmann:
do you want me to make a new PR? or can you delete the first commit here?

Simon-Steinmann:
okay I managed to fix it, now it's just the actual fix :)

DavidMansolino:
Do you have an example URDF that contains Lineset so that I can try?

Simon-Steinmann:
the kuka kr5 has some I think. it's in the automatic-conversion branch under automatic_conversion/urdf

.gitignore
```diff
@@ -2,3 +2,5 @@
 
 /urdf2webots.egg-info
 /dist
+automatic_conversion/protos/*
+automatic_conversion/ExtraProjectTest/*
```
DavidMansolino, line 6: This is not related to this PR.

urdf2webots/parserURDF.py
```diff
@@ -481,6 +481,9 @@ def getColladaMesh(filename, node, link):
     if hasattr(node, 'material') and node.material:
         for geometry in list(colladaMesh.scene.objects('geometry')):
             for data in list(geometry.primitives()):
+                if str(data.original).split()[0][1:] == 'LineSet':
```
DavidMansolino, line 484:
```suggestion
                if type(data.original) is lineset.LineSet:
```
This way the test is more self-understandable.
In addition, the collada import should be changed into: `from collada import Collada, lineset`

Simon-Steinmann: I'll test if that works. One additional note. Yesterday I saw, that Webots actually has lineSets as well. Perhaps instead of excluding them, we turn them into LineSets. Would require a few more lines off code, but should not be too hard.

DavidMansolino: Yes indeed, Webots does support IndexedLineSet: https://cyberbotics.com/doc/reference/indexedlineset
Supporting them completely in the converter, would indeed be the ideal solution. But in the meantime, merging this PR is not bad as it is already much better as it avoids having empty meshes.

Simon-Steinmann:
Done

urdf2webots/parserURDF.py
```diff
@@ -480,7 +481,7 @@ def getColladaMesh(filename, node, link):
     index = -1
     if hasattr(node, 'material') and node.material:
         for geometry in list(colladaMesh.scene.objects('geometry')):
-            for data in list(geometry.primitives()):
+            for data in list(geometry.primitives()):
```
DavidMansolino, line 484:
```suggestion
            for data in list(geometry.primitives()):
```

.gitignore
```diff
@@ -1,4 +0,0 @@
-*.pyc
-
-/urdf2webots.egg-info
-/dist
```
DavidMansolino, line 4: Why removing this file?

urdf2webots/writeProto.py
```diff
@@ -389,6 +390,8 @@ def URDFVisual(proto, visualNode, level, normal=False):
                 for value in visualNode.geometry.trimesh.coordIndex:
                     if len(value) == 3:
                         proto.write('%d %d %d -1 ' % (value[0], value[1], value[2]))
+                    if len(value) == 2:
```
DavidMansolino, line 393:
```suggestion
                    elif len(value) == 2:
                        assert visualNode.geometry.lineset
```

urdf2webots/writeProto.py
```diff
@@ -389,6 +390,8 @@ def URDFVisual(proto, visualNode, level, normal=False):
...
+                        proto.write('%d %d -1 ' % (value[0], value[1]))
             elif isinstance(visualNode.geometry.trimesh.coordIndex[0], np.int32):
                 for i in range(int(len(visualNode.geometry.trimesh.coordIndex) / 3)):
                     proto.write('%d %d %d -1 ' % (visualNode.geometry.trimesh.coordIndex[3 * i + 0],
```
DavidMansolino, line 397: Normal and texcoord (just below) have to be disabled in case of lineset.

DavidMansolino (changes requested):
That looks very good, just a few minor cleanups.

.gitignore
Simon-Steinmann: was done automatically

urdf2webots/writeProto.py
Simon-Steinmann: A LineSet should not have those, right?

DavidMansolino: Indeed lineset represents lines, there is not surface and therefore the normal and texture doesn't make sense (here is by the way the description of indexedLineSet: https://cyberbotics.com/doc/reference/indexedlineset).

Simon-Steinmann: This part should be resolved right?

DavidMansolino: Yes

.gitignore
DavidMansolino: I think it remains just to re-commit this file.

DavidMansolino (approved):
Looks al good now, thank you!

PR #81, opened 2020-09-09
Small inertia and toolslot w/ children fix

Simon-Steinmann:
The problems were:
1. that small inertia values were written as 0
2. If a urdf link specified as the tool-slot link had children (was the parent link of a joint), then it would format the generated proto incorrectly.


DavidMansolino (approved):
Thank you

PR #83, opened 2020-09-16
import .obj files as multiple meshes

Simon-Steinmann:
Imports .obj files more similarly to .dae files. Splitting them up into multiple meshes, if the .obj file contains multiple geometries.

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -421,15 +421,37 @@ def getSTLMesh(filename, node):
     return node
 
 
-def getOBJMesh(filename, node):
+def getOBJMesh(filename, node, link):
     """read obj file."""
-    print('Parsing Mesh: ' + filename)
+    print('Parsing Mesh: ' + filename)    
+    if hasattr(node, 'material') and node.material:
+        isVisual = True
+    else:
+        isVisual = False
+
     with open(filename, 'r') as file:
-        trimesh = node.geometry.trimesh
+        counter = 0
+        indexOffset = 0
         for line in file:
-            header, body = line.split(' ', 1)
+            header, body = line.split(' ', 1)            
             if header == '#':
                 continue
+            elif header == 'o': 
+                if counter != 0:
+                    if isVisual:
+                        link.visual.append(collision)
+                    else:
+                        link.collision.append(collision)
+                    indexOffset += len(collision.geometry.trimesh.coord)
+                if isVisual:
+                    collision = Visual()
+                else:
+                    collision = Collision()
+                collision.position = node.position
+                collision.rotation = node.rotation
+                collision.geometry.scale = node.geometry.scale
+                counter += 1
+                collision.geometry.name = '%s_%d' % (os.path.splitext(os.path.basename(filename))[0], counter)
             elif header == 'f':  # face
                 vertices = body.split()
                 coordIndex = []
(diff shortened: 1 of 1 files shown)


urdf2webots/parserURDF.py
```diff
@@ -421,53 +421,75 @@ def getSTLMesh(filename, node):
...
+    else:
+        isVisual = False
+
     with open(filename, 'r') as file:
-        trimesh = node.geometry.trimesh
+        counter = 0
+        indexOffset = 0
         for line in file:
-            header, body = line.split(' ', 1)
+            header, body = line.split(' ', 1)
```
DavidMansolino, line 436:
```suggestion
            header, body = line.split(' ', 1)
```

This kind of trailing spaces should be removed (it is by the way one of the reason of the test failure), I am sure you can configure your editor to remove them automatically when saving.

urdf2webots/parserURDF.py
```diff
@@ -421,52 +421,75 @@ def getSTLMesh(filename, node):
...
+        counter = 0
+        indexOffset = 0
         for line in file:
             header, body = line.split(' ', 1)
             if header == '#':
                 continue
+            elif header == 'o':
+                if counter != 0:
+                    if isVisual:
+                        link.visual.append(collision)
```
DavidMansolino, line 442: collison is not defined here.

DavidMansolino (commented):
It seems this PR breaks the DEF-USE mechanism for appearances:
```diff
diff --git a/tests/expected/Human.proto b/tests/expected/Human.proto
index 60380ae..1bfc48f 100644
--- a/tests/expected/Human.proto
+++ b/tests/expected/Human.proto
@@ -27,14 +27,14 @@ PROTO Human [
     selfCollision IS selfCollision
     children [
       Shape {
-        appearance DEF mat_pelvis_visual PBRAppearance {
+        appearance PBRAppearance {
           baseColor 0.500000 0.500000 0.500000
           transparency 0.000000
           roughness 1.000000
           metalness 0
           emissiveColor 0.000000 0.000000 0.000000
         }
-        geometry DEF Pelvis IndexedFaceSet {
+        geometry DEF Pelvis_1 IndexedFaceSet {
           coord Coordinate {
``

urdf2webots/parserURDF.py
Simon-Steinmann: I did a pep8 check, did not pick it up... hmmm

urdf2webots/parserURDF.py
Simon-Steinmann: that is the case for the first object, counter = 0, it will not append and create the collision object first. For the next 'o' it will then append it.

Simon-Steinmann:
> It seems this PR breaks the DEF-USE mechanism for appearances:
 will investigate

urdf2webots/parserURDF.py
DavidMansolino: Oh you're right, I was mislead by my linter (how also thought it was not defined).
Maybe this deserves a small comment, just to make it immediatly clear.

Simon-Steinmann: I can initialize it before to make pep8 happy :p

Simon-Steinmann:
Should be fixed now.

DavidMansolino (approved):
Thank you.

PR #86, opened 2020-09-18
Update Version

DavidMansolino:
Since the pip package was not updated since a few months and many recent improvement have been added to it, it would be nice to update it.


omichel (approved):
Thank you.

PR #90, opened 2020-10-01
added --insert-mesh --mesh-transform option to reduce file size of proto files. It also decreases loading times in webots significantly.

astumpf:
Please note that this is in an experimental state. It should provide a good starting point to officially and cleanly support this new feature.

Known Issues/Todos:
- STL and .obj import isn't tested yet
- Handling of collada files is tricky as they can consist of multiple parts stored in a single file. In general, this seems not to be well reflected in the urdf2webots importer. This patch prevents already multiple includes of the same mesh file when using the "--insert-mesh" option. However, in general, this is an open issue for all other cases, e.g. generating bounding box approximation results in heavily duplicated collision boxes in the protos file.
- Using collada files, an additional transform seems to be required. I'm not sure why this happens, but it seems that webots is not considering the collada coordinate system when loading a collada mesh from a file?
- Webots seems not to load subparts in the collada file correctly. It uses only the mesh, but not the stored appearance individually set for each part. Using the provided feature may lead to wrongly textured/colored meshes in favor of greatly reduced loading times.

Feel free to modify the code! :-)

Best regards

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -14,6 +14,11 @@ if __name__ == '__main__':
                          ' The filename minus the .proto extension will be the robot name.')
     optParser.add_option('--normal', dest='normal', action='store_true', default=False,
                          help='If set, the normals are exported if present in the URDF definition.')
+    optParser.add_option('--insert-mesh', dest='insertMesh', action='store_true', default=False,
+                         help='If set, the mesh is exported in the proto file. Otherwise the filepath to the meshes are exported.')
+    optParser.add_option('--mesh-transform', dest='meshTransform', default='',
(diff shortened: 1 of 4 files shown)


Simon-Steinmann:
Can you elaborate on what your code / feature does exactly?

astumpf:
I'm very sorry, but I'm currently very busy. Will definitely answer your request! Need some time to finish other very urgent tasks first.

omichel:
I am closing this PR for now as it has been inactive for a long period of time. Please re-open it if needed.

PR #91, opened 2020-10-06
Improve Test Robustness

DavidMansolino:
On some system the 'python' command is not available, but only 'python3'.


omichel (approved):
Nice.

PR #96, opened 2021-01-06
Fix package search without ROS1.

Simon-Steinmann:
Fixes the Path search, so a ROS1 repository does not need to be compiled and ROS1 does not need to be installed.

### DIFF
--- a/urdf2webots/importer.py
+++ b/urdf2webots/importer.py
@@ -75,7 +75,7 @@ def convert2urdf(inFile, outFile=None, normal=False, boxCollision=False,
             packageName = packages[0].split('/')[0]
             directory = os.path.dirname(inFile)
             while packageName != os.path.split(directory)[1] and os.path.split(directory)[1]:
-                directory = os.path.dirname(directory)
+                directory = os.path.dirname(os.path.abspath(inFile))
             if not os.path.split(directory)[1]:
                 try:
                     rospack = rospkg.RosPack()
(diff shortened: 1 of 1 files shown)


urdf2webots/importer.py
```diff
@@ -73,7 +73,7 @@ def convert2urdf(inFile, outFile=None, normal=False, boxCollision=False,
         packages = re.findall('"package://(.*)"', content)
         if packages:
             packageName = packages[0].split('/')[0]
-            directory = os.path.dirname(inFile)
+            directory = os.path.dirname(os.path.abspath(inFile))
```
omichel, line 76: It seems, that could be simplified:
```suggestion
            directory = inPath
```

omichel (changes requested):

urdf2webots/importer.py
Simon-Steinmann: inPath needs to stay unchanged. It would need to be a copy then. I had lots of troubles in the path with storing one variable in another and trying to keep on static. This commit might not be beautiful, but it works :)

omichel: `directory = inPath` actually makes a copy, because these variables are strings, not objects. So, it should work exactly as before.

Simon-Steinmann:
Ahh okay, good to know. I applied your suggestion.

omichel (approved):
Thank you.

Issue #97, opened 2021-01-14
Sensors are placed in base frame instead of respective frame


jgueldenstein:
Sensors (camera and IMU) are placed in the base frame instead of their respective frame.

This is the sensor description in our URDF:

```xml
<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="wolfgang">
[...]
  <!-- camera sensor -->
  <gazebo reference="camera">
      <sensor name='camera' type='camera'>
        <pose>-0.050000 -0.010000 0.000000 -1.5708 0.000000 -3.141592</pose>
        <camera name='__default__'>
          <horizontal_fov>0.57</horizontal_fov>
          <image>
            <width>620</width>
            <height>480</height>
          </image>
          <clip>
            <near>0.001000</near>
            <far>100.000000</far>
          </clip>
        </camera>
        <always_on>1</always_on>
        <update_rate>30.000000</update_rate>
        <visualize>1</visualize>
        <plugin name="camera_controller" filename="libgazebo_ros_camera.so">
          <alwaysOn>true</alwaysOn>
          <updateRate>30</updateRate>
          <cameraName>camera</cameraName>
          <imageTopicName>camera/image_proc</imageTopicName>
          <cameraInfoTopicName>camera_info</cameraInfoTopicName>
          <frameName>camera_optical_frame</frameName>
          <hackBaseline>0.07</hackBaseline>
          <distortionK1>0.0</distortionK1>
          <distortionK2>0.0</distortionK2>
          <distortionK3>0.0</distortionK3>
          <distortionT1>0.0</distortionT1>
          <distortionT2>0.0</distortionT2>
        </plugin>
      </sensor>
  </gazebo>
  [...]
  <!-- IMU sensor -->
  <gazebo reference="imu_frame">
      <gravity>true</gravity>
      <sensor name="imu_sensor" type="imu">
          <always_on>true</always_on>
          <update_rate>100</update_rate>
          <visualize>true</visualize>
          <topic>__default_topic__</topic>
          <plugin filename="libgazebo_ros_imu_sensor.so" name="imu_plugin">
              <topicName>imu</topicName>
              <bodyName>imu_frame</bodyName>
              <updateRateHZ>100.0</updateRateHZ>
              <gaussianNoise>0.0</gaussianNoise>
              <xyzOffset>0 0 0</xyzOffset>
              <rpyOffset>0 0 0</rpyOffset>
              <frameName>imu_frame</frameName>
          </plugin>
[...]

PR #98, opened 2021-01-14
interpret the reference attribute in gazebo tag as rootLink

jgueldenstein:
Fix to an issue with gazebo tags outside of links described in #97

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -923,7 +923,7 @@ def isRootLink(link, childList):
 
 def parseGazeboElement(element, parentLink, linkList):
     """Parse a Gazebo element."""
-    if element.hasAttribute("reference"):
+    if element.hasAttribute("reference") and any([link.name == element.getAttribute('reference') for link in linkList]):
         parentLink = element.getAttribute("reference")
     for plugin in element.getElementsByTagName('plugin'):
         if plugin.hasAttribute('filename') and plugin.getAttribute('filename').startswith('libgazebo_ros_imu'):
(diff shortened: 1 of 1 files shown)


urdf2webots/parserURDF.py
```diff
@@ -923,6 +923,8 @@ def isRootLink(link, childList):
 
 def parseGazeboElement(element, parentLink, linkList):
     """Parse a Gazebo element."""
+    if element.hasAttribute("reference"):
```
omichel, line 926: It seems this test was aldready done for the camera and lidar, but not for the IMU (with an extra condition):

```suggestion
    if element.hasAttribute("reference") and element.getAttribute('reference') in linkList:
```

We can them remove this test from lines 942-943 and 967-968.

omichel (changes requested):
Thank you for this contribution!

urdf2webots/parserURDF.py
jgueldenstein: The condition `element.getAttribute('reference') in linkList` does not work since the `linkList` contains `urdf2webots.parserURDF.Link` objects and the attribute is just a string. 
1. The behavior when the sensor is added to the sensor list but the parent link does not exist, is that the sensor is simply not added since https://github.com/cyberbotics/urdf2webots/blob/fe1a61eae935ffe5671e28fdb6be93209b4bbf68/urdf2webots/writeProto.py#L114 is never true.
2. The other option is to check if the link name exists in the list of link elements. If the link does not exist, the sensor would be placed in the base link.

So the code in 942-943 and 967-968 unfortunately never worked as intended.

Would you like to see option 1 or 2?

omichel: Thank you for the investigation and explanation. Option 2 looks better to me as it creates a useful list earlier.

jgueldenstein: The commit i made has this option implemented.

omichel (approved):
Very nice. Thank you!

PR #101, opened 2021-01-15
add dummy physics for imu sensor to make gyro work

jgueldenstein:
The gyro needs to be child to a Node with a a Physics element. Otherwise the following warning will pop up:
`WARNING: DEF Robot <robotname> (PROTO) > HingeJoint > Solid > HingeJoint > Solid > Solid > Gyro: this node or its parents requires a 'physics' field to be functional.`

In URDFs there is often an extra frame for an IMU which is attached by a fixed joint to the actual link the imu is attached to.

This PR adds a dummy physics node to the link holding the imu similar to the dummy physics added so tools don't fall off here: https://github.com/cyberbotics/urdf2webots/blob/aa6706adfc5c8039089aca0a330b54036d5386d4/urdf2webots/writeProto.py#L135

### DIFF
--- a/urdf2webots/writeProto.py
+++ b/urdf2webots/writeProto.py
@@ -2,7 +2,7 @@
 
 import math
 import numpy as np
-
+from urdf2webots.parserURDF import IMU
 from urdf2webots.math_utils import rotateVector, matrixFromRotation, multiplyMatrix, rotationFromMatrix
 
 toolSlot = None
@@ -112,6 +112,13 @@ def URDFLink(proto, link, level, parentList, childList, linkList, jointList, sen
         # 2: export Sensors
         for sensor in sensorList:
             if sensor.parentLink == link.name:
+                # add dummy physics for imu sensor if there is none since gyro requires it
+                if type(sensor) == IMU and link.inertia.mass is None:
+                    proto.write((level + 1) * indent + 'physics Physics {\n')
+                    proto.write((level + 1) * indent + '}\n')
+                    proto.write((level + 1) * indent + 'boundingObject Box {\n')
+                    proto.write((level + 1) * indent + 'size 0.01 0.01 0.01\n')
+                    proto.write((level + 1) * indent + '}\n')
                 if not haveChild:
                     haveChild = True
                     proto.write((level + 1) * indent + 'children [\n')


jgueldenstein:
The relevant part of the generated proto:
```
Solid {
  translation -0.025450 0.053500 -0.023500
  rotation -1.000000 -0.000000 -0.000000 1.570800
  physics Physics {
  }
  boundingObject Box {
  size 0.01 0.01 0.01
  }
  children [
    Accelerometer {
      name "imu_head accelerometer"
    }
    Gyro {
      name "imu_head gyro"
    }
    Compass {
      name "imu_head compass"
    }
  ]
  name "imu_frame_2"
}
```

omichel:
I am not sure this is the best way to address the problem. Adding a dummy physics-enable object will cause other problems, including performance drop and instability due to this useless mass. Instead, we should remove the intermediate Solid so that the Gyro is a direct child of the upper parent Solid which has physics, so instead, the relevant part should look like that:
```js
Accelerometer {
  translation -0.025450 0.053500 -0.023500
  rotation -1.000000 -0.000000 -0.000000 1.570800
  name "imu_head accelerometer"
}
Gyro {
  translation -0.025450 0.053500 -0.023500
  rotation -1.000000 -0.000000 -0.000000 1.570800
  name "imu_head gyro"
}
Compass {
  translation -0.025450 0.053500 -0.023500
  rotation -1.000000 -0.000000 -0.000000 1.570800
  name "imu_head compass"
}
```

jgueldenstein:
I agree and will work on it, a problem is if there are multiple fixed joints in series the first link/solid with physics when going up the kinematic chain towards the root should be used. The transforms would have to be multiplied.

omichel:
Note: Webots has a mechanism called Solid merger that avoids to create fixed joints. Instead, it multiplies transforms to merge them together into a single Solid, adding masses of each part. So, you should not worry about this in the PROTO file.

jgueldenstein:
I have tested the solution you proposed but it only works if there is a Physics node in the direct parent.

I think a better solution might be to make the following call recursive and adapt the transformation matrix. https://github.com/cyberbotics/webots/blob/a05248b57540c72651f7353dfd34718c8ffaceba/src/webots/nodes/WbGyro.cpp#L150

What do you think about this solution?

jgueldenstein:
Or somethink like here could be done: https://github.com/cyberbotics/webots/blob/a05248b57540c72651f7353dfd34718c8ffaceba/src/webots/nodes/WbAccelerometer.cpp#L154

omichel:
> Or somethink like here could be done: https://github.com/cyberbotics/webots/blob/a05248b57540c72651f7353dfd34718c8ffaceba/src/webots/nodes/WbAccelerometer.cpp#L154

Yes, this is exactly what I meant. I believe this is way to go.

StefanFabian:
Any update on this?
We're having the same problem and this PR is of great interest to us!

jgueldenstein:
We currently just add a dummy physics node to the Solid which has the IMU as children but I would like to fix this in webots in the future, I just have not gotten to it. It is a bit more complicated than I thought though since the acceleration is not the same as the first parent node with physics if there is a transform in between.

omichel:
@jgueldenstein: I am happy to help you if you have any question and to review your patch proposals.

PR #107, opened 2021-02-05
add option to use link names as DEF names

jgueldenstein:
To retrieve Nodes in the controller it is useful to use the function getFromProtoDef(<def_name>).

This pr adds an option to use the urdf link names as DEF names (in addition to solid names).


omichel (approved):
Thank you for this very nice contribution.

Issue #109, opened 2021-03-01
How to deal with controllers and sensors


reinzor:
Hi,

We're totally new to `webots` (we have been using `gazebo` so far) and we were wondering how to specify sensors and other plugins, e.g. controllers in the urdf. A lidar sensor in URDF (for gazebo) would be something like:

```
<gazebo reference="front_scan">
    <sensor name="front_scan" type="gpu_ray">
      <update_rate>34</update_rate>
      <pose>0 0 0 0 0 0</pose>
      <visualize>false</visualize>
      <ray>
        <scan>
          <horizontal>
            <samples>1072</samples>
            <resolution>1</resolution>
            <min_angle>-1.5533430342749535</min_angle>
            <max_angle>1.5533430342749535</max_angle>
          </horizontal>
        </scan>
        <range>
          <min>0.08</min>
          <max>30.0</max>
          <resolution>0.01</resolution>
        </range>
        <noise>
          <type>gaussian</type>
          <mean>0.0</mean>
          <stddev>0.001</stddev>
        </noise>
      </ray>
      <plugin filename="libgazebo_ros_gpu_laser.so" name="gazebo_ros_front_scan_controller">
        <topicName>front_scan</topicName>
        <frameName>front_scan</frameName>
      </plugin>
    </sensor>
  </gazebo>
```

How does this work for `webots`. We have tried the exporter and all links get exported (including geometry) but how to specificy sensors / actuators / plugins? We would like to maintain one robot description (like we have now).

Thanks!

jgueldenstein:
The issue is, that urdf2webots as of now only parses lidar sensors of `type="ray"`

I made a PR #111 to fix this.

reinzor:
@jgueldenstein, so this parser parses the gazebo tags and creates a proto model for this with the related sensors. But how do we point to custom plugins that we currently have. According to your documentation, this should be handled in the robot `controller`. How can we point to this controller and can we pass parameters via the urdf file to the controller?

jgueldenstein:
@reinzor In webots controllers work a bit different than sensor plugins in gazebo.
In webots you define your sensors and their properties such as field of view or resolution in the `.proto` file.
There are some existing controllers such as the webots_ros controller provide a service based interface to all the devices on your robot. You can also choose to write your own controller to keep the interface that you are using with the rest of your software or write a middle man software that translates between services and ros topics.

reinzor:
[...]

PR #111, opened 2021-03-01
handle gazebo sensor type gpu_ray same as ray

jgueldenstein:
`gpu_ray` seems to be a sensor type in gazebo sensor plugins. (see #109)

After this PR they are handled the same as a normal Lidar sensor.


jgueldenstein:
Attached a minimalistic urdf and the resulting proto file

[gpu_ray_test.zip](https://github.com/cyberbotics/urdf2webots/files/6064846/gpu_ray_test.zip)

omichel (approved):
Thank you!

PR #113, opened 2021-03-16
Fix typo in IMU parser

mael25:
There was a missing parenthesis that raised an error when converting a robot containing an IMU with non-zero Gaussian noise.


omichel (approved):
Well spotted!

PR #114, opened 2021-03-16
New workflow for CI test

mael25:
This workflow replaces the former .travis.yml test.


.github/workflows/travis.yml
```diff
@@ -0,0 +1,23 @@
+name: Test
```
lukicdarkoo, line 1: The file should be renamed to `test.yml`. Travis is the name of a CI that we don't want to rely on anymore.

.github/workflows/travis.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
```
lukicdarkoo, line 5:
```suggestion
    types: [opened, synchronize, reopened, labeled, unlabeled]
```
I don't like when the tests are restarted on `ready_for_review`, it is waste of time.

lukicdarkoo (approved):
Very good!

Make sure the `.travis.yml` file is deleted and that the unit tests are passing. You can test it faster by running them locally (`python3 -m unittest discover -s tests`).

mael25:
The workflow seems to do its job, but the test_export.py fails because it does not find the input URDF file (which is located on the [motorman repository](https://github.com/ros-industrial/motoman/tree/80c529979de52bb6eed5988d1e3076ae9abc131e) through a link):
`Input file "/home/runner/work/urdf2webots/urdf2webots/tests/sources/motoman/motoman_sia20d_support/urdf/sia20d.urdf" does not exists.`

I cloned the motoman repo at the right place and tried to run the tests locally. Then it finds the input and parses/creates meshes but fails to pass the test anyway:
`AssertionError: False is not true : Expected result mismatch when exporting "/home/mael/Desktop/urdf2webots/tests/sources/motoman/motoman_sia20d_support/urdf/sia20d.urdf"`

lukicdarkoo:
> The workflow seems to do its job, but the test_export.py fails because it does not find the input URDF file (which is located on the [motorman repository](https://github.com/ros-industrial/motoman/tree/80c529979de52bb6eed5988d1e3076ae9abc131e) through a link):

You are missing the `submodules` parameter, see here:
https://github.com/cyberbotics/webots_ros2/blob/d0f772b358252cda6b9e56768e5ec5cf7454a5ff/.github/workflows/test.yml#L18


> I cloned the motoman repo at the right place and tried to run the tests locally. Then it finds the input and parses/creates meshes but fails to pass the test anyway:
> `AssertionError: False is not true : Expected result mismatch when exporting "/home/mael/Desktop/urdf2webots/tests/sources/motoman/motoman_sia20d_support/urdf/sia20d.urdf"`

If check the merged PRs:
https://github.com/cyberbotics/urdf2webots/pulls?q=is%3Apr+is%3Aclosed
the tests haven't been working for a while. That means that the tests got broken at some point and we have to fix them :(

mael25:
The #VRML_SIM line was making the test_export.py fail because of a difference in the Webots version. Now, this line is skipped and if another line differs, both resulting and expected lines are displayed.

omichel (approved):
Thank you.
It looks good to me.
I removed the travis test from the branch protection settings and removed also the travis hooks from the settings.

lukicdarkoo (approved):
Great!

PR #115, opened 2021-03-19
Fix the conversion of URDF files including multiple package paths

schnaubelt:
In the current version, all package paths are replaced with the first path found.
However, we do have multiple robot URDFs that include components/sensor descriptions from multiple ROS packages, where this behavior breaks the mesh import. 
Using this PR, the importer replaces each 'package://' reference by the correct path.


lukicdarkoo:
@schnaubelt, thank you for the nice contribution!

Can you merge the `master` (tests are fixed there)?

I tried to push to your branch, but I am getting the 403 error. This is odd as I usually have enough permissions to push to forked branches.

schnaubelt:
@lukicdarkoo I just merged the master branch. Regarding the permissions: If you can recommend a specific setting change to allow this, I can change it.

lukicdarkoo (approved):
>  Regarding the permissions: If you can recommend a specific setting change to allow this, I can change it.

It is fine for this PR.

Thank you!

PR #118, opened 2021-04-16
Proposal: Fix handling of texture images

astumpf:
This merge request is intended for discussion about the current state of handling texture images. When texture images are used in combination with Collada mesh files, I encountered two issues that are addressed by this commit:

* When no tif file is provided, then the baseColor will be computed incorrectly later on (see [here](https://github.com/cyberbotics/urdf2webots/blob/48f0143d484dc6c9879ed11e437d1dc315fce657/urdf2webots/writeProto.py#L362)), which results in dark rendered images. This wrong behavior is fixed by setting visual.material.diffuse to a default value when a texture image is detected.

* The texture image coordinate system used by Webots differs from the commonly used system resulting in upside-down rendered images. This effect has been already described in an old [reference manual](http://edge.rit.edu/edge/P16201/public/P16201%20Google%20Drive%20Files/Tigerbot/P15201%20Documents/MSD/MSD%20I/WeBots_reference.pdf) in section 3.24.3. This commit includes the proposed transform by default.

Feedback is appreciated!

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -580,6 +580,9 @@ def getColladaMesh(filename, node, link):
                                             except IOError:
                                                 visual.material.texture = ""
                                                 print('failed to open ' + os.path.join(dirname, file))
+                            else:
+                                visual.material.diffuse = colorVector2Instance([1.0, 1.0, 1.0, 1.0])
+
                 link.visual.append(visual)
     else:
         for geometry in list(colladaMesh.scene.objects('geometry')):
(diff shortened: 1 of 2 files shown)


omichel:
I could review your contribution and it looks actually all good to me.
However, I would suggest that we also get the feedback from @lukicdarkoo who is our ROS2 expert (he is currently off and will be back next week on Tuesday).

lukicdarkoo:
Looks good to me as well.

I am only not sure about the default color you chose, it is specified by Collada? In Webots, it is [0.8, 0.8, 0.8, 0.8]:
https://cyberbotics.com/doc/reference/material

astumpf:
I will try it and let you know.

astumpf:
The default color [0.8, 0.8, 0.8, 0.8] let appear my images "greyish". Here a comparison:

Using [0.8, 0.8, 0.8, 0.8]:
Screenshot from 2021-05-19 05-34-56

Using [1.0, 1.0, 1.0, 1.0]:
Screenshot from 2021-05-19 05-36-52

omichel (approved):
That looks good to me.

lukicdarkoo (approved):
It looks good to me as well, thank you

PR #119, opened 2021-04-26
Implemented P3D object for ground truth pose output by Webots

astumpf:
Adds P3D object that is created when parsing a "libgazebo_ros_p3d" from the URDF file. Aftward, all required Nodes (GPS, InertialUnit and Gyro) are added to the referred link, which then can be used, e.g. to implement a ground truth publisher node.

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -991,6 +991,10 @@ def parseGazeboElement(element, parentLink, linkList):
             p3d.parentLink = parentLink
             if hasElement(plugin, 'topicName'):
                 p3d.name = plugin.getElementsByTagName('topicName')[0].firstChild.nodeValue
+            if hasElement(plugin, "xyzOffsets"):
+                print('\033[1;33mWarning: URDF parser cannot handle \"xyzOffsets\" from p3d!\033[0m')
+            if hasElement(plugin, "rpyOffsets"):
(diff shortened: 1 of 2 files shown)


urdf2webots/parserURDF.py
```diff
@@ -944,6 +986,12 @@ def parseGazeboElement(element, parentLink, linkList):
                     if link.name == name:
                         link.forceSensor = True
                         break
+        elif plugin.hasAttribute('filename') and plugin.getAttribute('filename').startswith('libgazebo_ros_p3d'):
+            p3d = P3D()
```
lukicdarkoo, line 990: Maybe a warning here to notify a user that parameters `xyzOffsets` and `rpyOffsets` cannot be handled

astumpf: Yes, that is something I'm also concerned about. I guess you could also add a Transform node to handle this, but I got only little time to work on these changes. Therefore, I would suggest to add a kind of "not implemented yet" warning?

lukicdarkoo: A warning should be enough. If the `xyzOffsets` or `rpyOffsets` parameter is present we should just print a warning.

astumpf: The latest commit includes the suggested changes. A warning is now thrown when either `xyzOffsets` or `rpyOffsets` parameter is present.

lukicdarkoo (approved):
Thank you

PR #124, opened 2021-07-16
Update version to R2021b

BenjaminDeleze:
update to R2021b


lukicdarkoo:
Maybe we should also upgrade the headers in the `tests/expected` directory?

BenjaminDeleze:
Good idea, it seems that the tests were already broken, I will try to fix them

BenjaminDeleze:
It should be good now. I will fix the release info page too.

lukicdarkoo (approved):
Thank you

PR #128, opened 2021-08-20
Feature lidar transformation and rangefinder

Highfly1212:
1. Added the range finder class, that reads in a depth camera from gazebo and parses it as a range finder node to webots.
2. Added the clip field to the lidar to fill in the near field in webots.
3. Added rotation in the lidar node (first rotation -90 degrees along x-axis, then 90 degrees along z-axis), to fit ros conventions (x-axis to the front of the lidar, z-axis up, in case of a rotating lidar, rotating around z-axis).


lukicdarkoo:
@Highfly1212 Thank you for your contribution!

However, the range finder seems not to be properly oriented:
Screenshot-20210823145941-660x595

I tested it with the URDF/XACRO provided in this package:
https://github.com/nilseuropa/realsense_ros_gazebo

Highfly1212:
@lukicdarkoo thanks for your answer. I just tested the code with the official realsense description files from Intel (from https://github.com/IntelRealSense/realsense-ros/tree/development/realsense2_description). 
Only added a gazebo file for the sensors, I suspect that the parser doesn't work with the package you used for testing, because it doesn't include a reference for each sensor in the depthcam.xacro file. Instead it uses gazebo specific tags like  <depthOpticalframeName>. I suspect they get ignored by the parser. 
 A picture of the rangefinder are attached. The gazebo file looks as following:

<?xml version="1.0"?>
<robot xmlns:xacro="http://ros.org/wiki/xacro">

  <xacro:arg name="control_rate" default="100" />

  
    <gazebo reference="camera_depth_optical_frame">
      <sensor type="depth" name="d435_depth_camera">
        <update_rate>30</update_rate>
        <camera>
          <horizontal_fov>${87.0 * pi/180.0}</horizontal_fov>
          <image>
            <format>B8G8R8</format>
            <width>1280</width>
            <height>720</height>
          </image>
        </camera>
        <range>
          <min>0.0</min>
          <max>2.0</max>
        </range>
      </sensor>
    </gazebo>

    
    <gazebo reference="camera_color_optical_frame">
      <sensor type="camera" name="d435_color_camera">
        <pose>0 0 0 0 0 0</pose>
        <update_rate>30.0</update_rate>
        <camera>
          <horizontal_fov>${87.0 * pi/180.0}</horizontal_fov>
          <image>
            <width>1280</width>
            <height>720</height>
            <format>R8G8B8</format>
          </image>
          <clip>
            <near>0.0</near>
            <far>20.0</far>
          </clip>
        </camera>
      </sensor>
    </gazebo>
</robot>
realsense_d435_rangefinder

lukicdarkoo:
I see, thank you for the explanation.

However, I am afraid that might be the issue. Comparing to the `realsense_ros_gazebo` package, the Gazebo camera is linked to the `camera_depth_optical_frame` instead of the `camera_link`. Have you tried importing the URDF to Gazebo?

I am not super experienced in this area, so I might be wrong. However, it is critical the orientation to be correct, especially since we are doing a huge refactoring (https://github.com/cyberbotics/webots/pull/3073, the new camera is oriented done according to the `realsense_ros_gazebo` package). There is even an analysis of the camera orientation https://github.com/cyberbotics/webots/pull/3472#issuecomment-888181072

Highfly1212:
Hey, 
I just checked with my d455 to see which frames the images in the ROS messages are assigned to. The color camera is assigned to camera_color_optical_frame and the depth camera to camera_depth_optical_frame (Which can be seen in the screenshots).
 Additionally I added a picture from the d435 in Webots, showing that both the camera and the rangefinder are oriented correctly. I suspect that with your test packages the camera should also be incorrectly oriented like the rangefinder, as they have the same rotation (could you please check this?).  The camera is already oriented that way in the parser, I just added the rangefinder accordingly. 
In the realsense_ros_gazebo package the color camera as well as the depth camera get attached to the camera_link, but later on they are attached to the optical frame with the tags "colorOpticalframeName" respectively "depthOpticalframeName" (which you can see in https://github.com/nilseuropa/realsense_ros_gazebo/blob/master/xacro/depthcam.xacro#L233). According to my knowledge the parser neither considers these tags nor the plugins in general for all sensor elements. So for it to work the gazebo reference must be set on the optical frames. 
ros_msg_color_camera
ros_msg_depth_camera
realsense_d435_rangefinder_color_camera

lukicdarkoo:
I converted the above URDF model to PROTO. Then, I position both "robots" to look at the box. And here is what we get:

Screenshot-20210908111000-710x516

Screenshot-20210908111202-506x338

To me, it seems to be wrong. I don't where does the (pi, 0, 0) rotation comes from.

lukicdarkoo:
@omichel Could you please take a look into this?

astumpf:
Closing this request and resubmitting a new version in multiple request for better review.

PR #130, opened 2021-08-23
Fix root link search

### LINKED ISSUE #129: Converting the `realsense_ros_gazebo` URDF outputs almost empty PROTO
I tried to export a PROTO from the [`nilseuropa/realsense_ros_gazebo`](https://github.com/nilseuropa/realsense_ros_gazebo) but it doesn't export any meshes.

I believe the problem is here:
https://github.com/cyberbotics/urdf2webots/blob/d4c366bad0de52c4cfa44792a2e74aba6ee99c97/urdf2webots/importer.py#L153-L154

as `d435_bottom_screw_frame` and `d435_link` links are not exported. I believe `if root link has only one joint which type is fixed, it should not be part of the model` is a very strong assumption and we should better handle it.

lukicdarkoo:
Fixes #129

I am not sure whether this is the best fix, but it works for #129 and it is passing the tests.


omichel (approved):
That seems pretty good to me.

PR #133, opened 2021-10-18
Link mesh files for visual and collision instead of creating indexed face sets

BenjaminHug8:
This PR is for linking the mesh files instead of transforming them into index face sets when converting URDF to PROTO, thus the PROTO files are way more shorter and Webots can load faster.


BenjaminHug8:
There is code in commentary that I will delete once you approved (maybe should we keep it) ?

BenjaminHug8:
I just noticed one thing in Human.proto, the closing `}` is not correctly indented, I am on it.

omichel:
> There is code in commentary that I will delete once you approved (maybe should we keep it) ?

Yes, we should delete it.

BenjaminHug8:
Ok I removed the code in commentaries and check once the code, so this PR should be ok now.

I let you @omichel do a final check with the review.

urdf2webots/parserURDF.py
```diff
@@ -87,6 +88,14 @@ def __init__(self):
         self.radius = 0.0
 
 
+class Mesh():
+    """Define mesh object."""
+
+    def __init__(self):
+        """Initializatization."""
+        self.url = ""
```
omichel, line 96: We prefer single quotes:
```suggestion
        self.url = ''
```

urdf2webots/parserURDF.py
```diff
@@ -848,17 +642,12 @@ def getCollision(link, node, path):
...
-                getOBJMesh(meshfile, collision, link)
-            elif extension == '.stl':
+            if extension in extensionList:
                 name = os.path.splitext(os.path.basename(meshfile))[0]
                 if name in Geometry.reference:
                     collision.geometry = Geometry.reference[name]
                 else:
-                    if extension == '.stl':
-                        collision.geometry.stl = getSTLMesh(meshfile, collision)
+                    collision.geometry.mesh.url = "\"" + meshfile + "\""
```
omichel, line 650:
```suggestion
                    collision.geometry.mesh.url = '"' + meshfile + '"'
```

omichel (changes requested):
It looks good to me. Only a few minor things.

urdf2webots/parserURDF.py
```diff
@@ -786,17 +585,12 @@ def getVisual(link, node, path):
...
-                getOBJMesh(meshfile, visual, link)
-            elif extension == '.stl':
+            if extension in extensionList:
                 name = os.path.splitext(os.path.basename(meshfile))[0]
                 if name in Geometry.reference:
                     visual.geometry = Geometry.reference[name]
                 else:
-                    if extension == '.stl':
-                        visual = getSTLMesh(meshfile, visual)
+                    visual.geometry.mesh.url = "\"" + meshfile + "\""
```
omichel, line 593: Missed that one:
```suggestion
                    visual.geometry.mesh.url = '"' + meshfile + '"'
```

urdf2webots/writeProto.py
```diff
@@ -397,93 +376,20 @@ def URDFVisual(proto, visualNode, level, normal=False):
...
-                        proto.write('%d %d %d -1 ' % (value[0], value[1], value[2]))
-                    elif len(value) == 2:
-                        assert visualNode.geometry.lineset
-                        proto.write('%d %d -1 ' % (value[0], value[1]))
-            elif isinstance(visualNode.geometry.trimesh.coordIndex[0], np.int32):
-                for i in range(int(len(visualNode.geometry.trimesh.coordIndex) / 3)):
-                    proto.write('%d %d %d -1 ' % (visualNode.geometry.trimesh.coordIndex[3 * i + 0],
-                                                  visualNode.geometry.trimesh.coordIndex[3 * i + 1],
-                                                  visualNode.geometry.trimesh.coordIndex[3 * i + 2]))
+                proto.write((shapeLevel + 1) * indent + 'geometry DEF %s Mesh {\n' % (visualNode.geometry.defName))
```
omichel, line 386: I believe these extra parenthesis are not needed:
```suggestion
                proto.write((shapeLevel + 1) * indent + 'geometry DEF %s Mesh {\n' % visualNode.geometry.defName)
```

omichel (changes requested):

omichel:
After this change, it seems that the following options are now obsolete: `--disable-mesh-optimization` `--multi-file`. We should probably remove them and update the Readme.md accordingly.

BenjaminHug8:
> After this change, it seems that the following options are now obsolete: `--disable-mesh-optimization` `--multi-file`. We should probably remove them and update the Readme.md accordingly.

I also remove `Trimesh` as it was used before for the conversion of mesh into index face set.

urdf2webots/writeProto.py
```diff
@@ -7,7 +7,6 @@
 
 toolSlot = None
 staticBase = False
-enableMultiFile = False
 meshFilesPath = None
```
omichel, line 10: This one seems to be useless now:
```suggestion
```

omichel (changes requested):

BenjaminHug8:
Should I update the `Readme.md` to state that we removed the 2 options and that now we are directly linking the meshes instead or is it not necessary ?

omichel:
I believe we should introduce a CHANGELOG.md file at the top level and start logging this kind of changes under version numbers.

omichel:
> I believe we should introduce a CHANGELOG.md file at the top level and start logging this kind of changes under version numbers.

No, that's not a good idea as it seems there is already some automatic generation of change log, see the [releases page](https://github.com/cyberbotics/urdf2webots/releases).
So we should simply ensure this automatic change log contains clear information about the change (thus the title of this PR is probably super important).

BenjaminHug8:
> > I believe we should introduce a CHANGELOG.md file at the top level and start logging this kind of changes under version numbers.
> 
> No, that's not a good idea as it seems there is already some automatic generation of change log, see the [releases page](https://github.com/cyberbotics/urdf2webots/releases). So we should simply ensure this automatic change log contains clear information about the change (thus the title of this PR is probably super important).

So can we just have to merge this PR and prepare for a new release ?
I am just wondering why we get from version `1.0.9` to `10.0.10` ?

omichel:
> So can we just have to merge this PR and prepare for a new release?

Yes. That's probably a new GitHub feature that was not available before 10.0.10 or that we didn't enable when making the previous releases.

omichel (approved):
Thank you.

PR #136, opened 2021-11-09
conversion enu/flu

BenjaminHug8:
This PR prepare the conversion enu/flu that will be introduce in `Webots R2022a`.

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -37,7 +37,7 @@ class Inertia():
     def __init__(self):
         """Initializatization."""
         self.position = [0.0, 0.0, 0.0]
-        self.rotation = [1.0, 0.0, 0.0, 0.0]
+        self.rotation = [0.0, 0.0, 1.0, 0.0]
         self.mass = None
         self.ixx = 1.0
         self.ixy = 0.0
@@ -150,7 +150,7 @@ class Visual():
     def __init__(self):
         """Initializatization."""
         self.position = [0.0, 0.0, 0.0]
-        self.rotation = [1.0, 0.0, 0.0, 0.0]
+        self.rotation = [0.0, 0.0, 1.0, 0.0]
         self.geometry = Geometry()
         self.material = Material()
 
@@ -161,7 +161,7 @@ class Collision():
     def __init__(self):
         """Initializatization."""
         self.position = [0.0, 0.0, 0.0]
-        self.rotation = [1.0, 0.0, 0.0, 0.0]
+        self.rotation = [0.0, 0.0, 1.0, 0.0]
         self.geometry = Geometry()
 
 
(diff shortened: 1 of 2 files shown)


joachimhgg (changes requested):
I tried to convert the kuka following this tutorial: https://github.com/cyberbotics/urdf2webots/blob/master/docs/tutorial.md but I got this result:
image

We also need to update the tutorial.

BenjaminHug8:
> I tried to convert the kuka following this tutorial: https://github.com/cyberbotics/urdf2webots/blob/master/docs/tutorial.md but I got this result

It does not work in `master` either...
I am going to check if that is coming from the new linking of the meshes.

BenjaminHug8:
Ok it seems that this kuka arm is missing some rotation in the URDF thus also the XACRO compared to the test on CI with motoman.

BenjaminHug8:
Up to version 1.0.10 it works correctly.

The base missing is not surprising as it is defined with an indexed face set that I removed when linking the meshes.

But I have to solve the orientation of each part and the textures missing, thanks for spotting this!

BenjaminHug8:
> I have to solve the orientation of each part and the textures missing, thanks for spotting this!

This will be done in other PR.

BenjaminHug8:
> I tried to convert the kuka following this tutorial: https://github.com/cyberbotics/urdf2webots/blob/master/docs/tutorial.md but I got this result: image

This has been solved in another PR (the mesh linking was not working properly for `.dae` files).

urdf2webots/importer.py
```diff
@@ -208,7 +208,7 @@ def convert2urdf(inFile, outFile=None, normal=False, boxCollision=False,
                          help='If set, the base link will have the option to be static (disable physics)')
     optParser.add_option('--tool-slot', dest='toolSlot', default=None,
                          help='Specify the link that you want to add a tool slot too (exact link name from urdf)')
-    optParser.add_option('--rotation', dest='initRotation', default='0 1 0 0',
+    optParser.add_option('--rotation', dest='initRotation', default='0 0 1 0',
                          help='Set the rotation field of your PROTO file.')
     optParser.add_option('--init-pos', dest='initPos', default=None,
                          help='Set the initial positions of your robot joints. Example: --init-pos="[1.2, 0.5, -1.5]" would '
```
joachimhgg, line 214: maybe the example of the initial-pos needs to be adapt

joachimhgg (changes requested):
The images need to be update. (I don't have the meshes folder anymore, etc)
image
image

urdf2webots/importer.py
BenjaminHug8: This is an option, it is not related to one example in particular.

BenjaminHug8:
Images changes.

joachimhgg (approved):
Thanks!

omichel (approved):

BenjaminHug8:
@omichel I ask you to review once again as I prepared the `1.0.13` release and removed the python 2 test.

I removed some white spaces at the end of some lines added during the creation of the PROTO.
There is also extra space and comma after the last item of some list but I kept them as it would take a bit more place in the code to only to polish a huge list of point. In my opinion it is not worth to do it but let me know if you think it is a good idea to remove these extra characters.

omichel (approved):
Looks good to me. Thank you.

PR #139, opened 2021-11-16
Remove the mesh linking for dae files

BenjaminHug8:
This PR is to revert the mesh linking for `.dae` files as they might have multiple material colors. This is currently not supported by the `Mesh` node of Webots.


urdf2webots/parserURDF.py
```diff
@@ -28,9 +28,23 @@
 
 # to pass from external
 robotName = ''
-extensionList = ['.3ds', '.blend', '.bvh', '.dae', '.fbx', '.stl', '.obj', '.x3d']
+extensionListSingleColor = ['.stl', '.obj']
```
omichel, line 31: I would rename this variable to make it clearer:
```suggestion
extensionListSingleAppearance = ['.stl', '.obj']
```

BenjaminHug8:
Oh sorry I was sure that this PR was in `draft`.
I have to correct the tests.

BenjaminHug8:
I also confirm that the most standard format for the mesh files used in URDF are `.stl`, `.dae` and `.obj`.

It has been rarely mentioned that `.fbx` and `.mesh` can also be used in URDF files, but they seem to be almost never used by the community.

BenjaminHug8:
@omichel is it ok for you if after the merging of this PR I overwrite the version `1.0.11` as `.dae` conversion is not fully functional now?

omichel (approved):
I would like that @lukicdarkoo also review this PR before we merge it.

lukicdarkoo (commented):
Looks good.

Could you add tests? It would be nice to have tests for .dae files, just to be able to quickly verify whether the .dae conversion works as expected.

BenjaminHug8:
Addition of a test on a kuka arm with `.dae` and `.stl` mesh files.

Correction of conversion of `indexLineSet` to use `Appearance` instead of `PBRAppearance`.

lukicdarkoo (approved):
Perfect, thanks!

omichel (approved):

PR #142, opened 2021-12-06
Import URDF files as robots inside world files

BenjaminHug8:
In order to use the URDF importer on the fly with `webots_ros2` we have to change the way how `urdf2webots` works.

Instead of creating a PROTO file and then manually add it in the world file the tool will now convert the URDF file into a robot and insert it in the world file.


BenjaminHug8:
I close this in favor of another PR as we changed the workflow.
The new PR will be done with a branch checkout from `develop`.

BenjaminHug8:
PR #146 is the new PR replacing this one.

PR #147, opened 2022-01-21
Remove special rotation for cylinders

BenjaminHug8:
Some URDF files were not correctly converted, as some links were not rotated in the correct way. This was due to a special case for the cylinder before the conversion to FLU/ENU.

This PR remove this exception.

### DIFF
--- a/urdf2webots/math_utils.py
+++ b/urdf2webots/math_utils.py
@@ -26,8 +26,8 @@ def rotationFromQuaternion(q):
     if (v[3] < 0.0001):
         # if v[3] close to zero then direction of axis not important
         v[0] = 0.0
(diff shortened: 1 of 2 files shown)


omichel (approved):
Well spotted!

Timple:
Would it be possible to create a new python-pip release with this fix?

BenjaminHug8:
> Would it be possible to create a new python-pip release with this fix?

It should be out with [1.0.14](https://github.com/cyberbotics/urdf2webots/releases/tag/1.0.14).
Do you still have issues with 1.0.14 and beyond for the cylinders?

Timple:
Ah, missed that. Thanks!

PR #155, opened 2022-02-07
Fix runtime error when parsing meshes due to using float in range loop

astumpf:
Simple fix of runtime error when parsing meshes.

### DIFF
--- a/urdf2webots/writeRobot.py
+++ b/urdf2webots/writeRobot.py
@@ -548,7 +548,7 @@ def URDFVisual(robotFile, visualNode, level, normal=False):
                         if len(value) == 3:
                             robotFile.write('%d %d %d -1 ' % (value[0], value[1], value[2]))
                 elif isinstance(visualNode.geometry.trimesh.texCoordIndex[0], np.int32):
-                    for i in range(len(visualNode.geometry.trimesh.texCoordIndex) / 3):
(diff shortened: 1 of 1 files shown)


omichel:
Do you have an example of a mesh which was causing trouble here? Normally the length of the texture coordinate index list should always be a multiple of 3, otherwise, your mesh may be corrupted.

astumpf:
It is indeed always 3, but division returns a `float` type, which cannot be used as `range`.

urdf2webots/writeRobot.py
```diff
@@ -548,7 +548,7 @@ def URDFVisual(robotFile, visualNode, level, normal=False):
                         if len(value) == 3:
                             robotFile.write('%d %d %d -1 ' % (value[0], value[1], value[2]))
                 elif isinstance(visualNode.geometry.trimesh.texCoordIndex[0], np.int32):
-                    for i in range(len(visualNode.geometry.trimesh.texCoordIndex) / 3):
+                    for i in range(int(len(visualNode.geometry.trimesh.texCoordIndex) / 3)):
```
omichel, line 551: I see, in that case, we could maybe use the `//` operator:
```suggestion
                    for i in range(len(visualNode.geometry.trimesh.texCoordIndex) // 3):
```

astumpf: You're right. Wasn't aware of the integer division.

omichel (approved):

PR #156, opened 2022-02-07
Removing image coordinate flipping

astumpf:
I think I've seen also a changelog addressing changes in Webots coordinate systems, but cannot find it anymore. However, the coordinate flipping for textures isn't necessary anymore using the latest version of Webots.

### DIFF
--- a/urdf2webots/writeRobot.py
+++ b/urdf2webots/writeRobot.py
@@ -449,7 +449,7 @@ def URDFVisual(robotFile, visualNode, level, normal=False):
                 robotFile.write((shapeLevel + 2) * indent + '}\n')
 
                 robotFile.write((shapeLevel + 2) * indent + 'textureTransform TextureTransform {\n')
-                robotFile.write((shapeLevel + 3) * indent + 'scale 1 -1\n')
+                robotFile.write((shapeLevel + 3) * indent + 'scale 1 1\n')
                 robotFile.write((shapeLevel + 2) * indent + '}\n')
             robotFile.write((shapeLevel + 1) * indent + '}\n')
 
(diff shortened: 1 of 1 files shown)


urdf2webots/writeRobot.py
```diff
@@ -449,7 +449,7 @@ def URDFVisual(robotFile, visualNode, level, normal=False):
                 robotFile.write((shapeLevel + 2) * indent + '}\n')
 
                 robotFile.write((shapeLevel + 2) * indent + 'textureTransform TextureTransform {\n')
-                robotFile.write((shapeLevel + 3) * indent + 'scale 1 -1\n')
+                robotFile.write((shapeLevel + 3) * indent + 'scale 1 1\n')
```
omichel, line 452: Since `1 1` is the default value for scale, it could simply be dropped:
```suggestion
```

BenjaminHug8, line 452: As `scale` has the default value, a better fix would be to remove the three lines adding the `textureTransform` node.

BenjaminHug8 (changes requested):
You are indeed correct, I just have a minor request.

urdf2webots/writeRobot.py
BenjaminHug8: Well @omichel has been faster than me.

astumpf: As suggested by BenjaminHug8 I've dropped the TextureTransform node.

omichel: But your point is better than mine as the `TextureTransform` node is actually useless.

omichel (approved):

BenjaminHug8 (approved):
Thank you!

Issue #153, opened 2022-02-04
Single appearance feature for meshfiles overwrites mesh scaling


astumpf:
In example:
https://github.com/cyberbotics/urdf2webots/blob/f0a990cccdc28641db3c38639fd6f5d7a8e150fd/urdf2webots/parserURDF.py#L680

This causes the visual to lose its individual properties such as transformation. The assumption that the same meshfile will be re-used with the identical transformation is wrong.

BenjaminHug8:
Thank you for the report, I will take a look.

BenjaminHug8:
I just tested and everything is fine on my side, I can use two links with the same mesh and have to joints with different origins and the results is ok.

The visual object in our tool contains a translation and rotation:
https://github.com/cyberbotics/urdf2webots/blob/f0a990cccdc28641db3c38639fd6f5d7a8e150fd/urdf2webots/parserURDF.py#L159-L167
and as you can see the geometry is variable of this class.

The line you mentioned is just for reusing the same `Mesh` node contained in the geometry, but that's all.
Let me know if you still have questions or if your issue is solved or not.

astumpf:
Yes, you're right. I overlooked this detail.  To give full detail about my concern: I'm working on a patch to consider also scaling of meshes. But the scale is indeed part of the `Geometry` object (https://github.com/cyberbotics/urdf2webots/blob/f0a990cccdc28641db3c38639fd6f5d7a8e150fd/urdf2webots/parserURDF.py#L159-L167) that will be overwritten as stated above.

BenjaminHug8:
> However, this scale patch would become obsolete in my case, when #152 was solved.

Yes exactly, I am implementing the `scale` option of `mesh` tags for solving this issue.

BenjaminHug8:
@astumpf the PR #157 has been merged and should solve your issue. Feel free to reopen this issue if needed.

astumpf:
Thanks a lot!

Issue #152, opened 2022-02-04
How to reflect meshes loaded from files?


astumpf:
Robots are usually symmetrical and therefore many parts are identical but reflected. Using URDF it's common practice to just use the identical mesh multiple times and just reflect it by using negative scales, e.g. for mirroring along the x-z plane (sagittal) you would use the scaling vector [1 -1 1]. However, Webots forbids this type of mesh reflection using the Transform node (https://cyberbotics.com/doc/reference/transform).

```
Transform {
  translation 0.000000 0.000000 0.060000
  rotation 0.000000 0.000000 1.000000 0.000000
  scale 1.000000 -1.000000 1.000000
  children [
    Shape {
      appearance PBRAppearance {
        baseColor 0.500000 0.500000 0.500000
        transparency 0.000000
        roughness 1.000000
        metalness 0
        emissiveColor 0.000000 0.000000 0.000000
      }
      geometry DEF my_mesh Mesh {
        url "my_awesome_mesh.obj"
      }
    }
  ]
  }
```

Is there any way to reuse a single mesh file in case of reflections?

BenjaminHug8:
Hi @astumpf,

As you have maybe noticed from the doc you mentioned, scale cannot be negative in Webots.

However I think we should be able to add boolean in the `Mesh` node to mirrors the mesh along a certain plane. I just need to look how URDF specifies this mirrors effects to know how we could use it in the conversion tool.

Thank you for the proposition!

astumpf:
Sounds good! I am already looking forward to the update.

BenjaminHug8:
@astumpf the PR #157 has been merged in the `develop` branch and should solve this issue. Feel free to reopen if needed.

PR #157, opened 2022-02-08
Add the scale field to Mesh and Collada nodes

BenjaminHug8:
This PR solves #152 and #153.

Linked to https://github.com/cyberbotics/webots/pull/4243.

### DIFF
--- a/urdf2webots/writeRobot.py
+++ b/urdf2webots/writeRobot.py
@@ -283,6 +283,10 @@ def URDFBoundingObject(robotFile, link, level, boxCollision):
                     robotFile.write(initialIndent + 'Mesh {\n')
 
                 robotFile.write((boundingLevel + 1) * indent + 'url ' + str(boundingObject.geometry.mesh.url) + '\n')
+                if boundingObject.geometry.scale[0]:
+                    robotFile.write((boundingLevel + 1) * indent + 'scale %lf %lf %lf\n' % (boundingObject.geometry.scale[0],
+                                                                                      boundingObject.geometry.scale[1],
+                                                                                      boundingObject.geometry.scale[2]))
                 robotFile.write(boundingLevel * indent + '}\n')
 
         else:
@@ -324,6 +328,10 @@ def URDFVisual(robotFile, visualNode, level, normal=False):
                 robotFile.write(shapeLevel * indent + 'ColladaShapes {\n')
 
             robotFile.write((shapeLevel + 1) * indent + 'url ' + str(visualNode.geometry.colladaShapes.url) + '\n')
+            if visualNode.geometry.scale[0]:
+                    robotFile.write((shapeLevel + 1) * indent + 'scale %lf %lf %lf\n' % (visualNode.geometry.scale[0],
+                                                                                      visualNode.geometry.scale[1],
+                                                                                      visualNode.geometry.scale[2]))
     else:
         robotFile.write(shapeLevel * indent + 'Shape {\n')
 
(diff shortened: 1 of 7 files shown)


urdf2webots/writeRobot.py
```diff
@@ -239,15 +239,18 @@ def URDFBoundingObject(robotFile, link, level, boxCollision):
...
             robotFile.write((boundingLevel + 1) * indent + 'rotation %lf %lf %lf %lf\n' % (boundingObject.rotation[0],
-                                                                                       boundingObject.rotation[1],
-                                                                                       boundingObject.rotation[2],
-                                                                                       boundingObject.rotation[3]))
+                                                                                        boundingObject.rotation[1],
+                                                                                        boundingObject.rotation[2],
+                                                                                        boundingObject.rotation[3]))
+            robotFile.write((boundingLevel + 1) * indent + 'scale %lf %lf %lf\n' % (boundingObject.scale[0],
+                                                                                    boundingObject.scale[1],
+                                                                                    boundingObject.scale[2]))
```
omichel, line 253: I have a general remark here: if the value to be written for a translation, rotation or scale (or any other field) is the default value, then we shouldn't write it. That would make the proto output much cleaner and more readable.

omichel (approved):
That seems good to me.
I know my suggestion is not the purpose of this PR, but we could make it here if easy to do or in a separate PR.

urdf2webots/writeRobot.py
BenjaminHug8: Totally agree with that. I will do it here.

urdf2webots/writeRobot.py
```diff
@@ -260,10 +268,10 @@ def URDFBoundingObject(robotFile, link, level, boxCollision):
                                                                                 boundingObject.geometry.box.z))
             robotFile.write(boundingLevel * indent + '}\n')
 
-        elif boundingObject.geometry.cylinder.radius != 0 and boundingObject.geometry.cylinder.length != 0:
+        elif boundingObject.geometry.cylinder.radius != 0 and boundingObject.geometry.cylinder.height != 0:
             robotFile.write(initialIndent + 'Cylinder {\n')
             robotFile.write((boundingLevel + 1) * indent + 'radius ' + str(boundingObject.geometry.cylinder.radius) + '\n')
-            robotFile.write((boundingLevel + 1) * indent + 'height ' + str(boundingObject.geometry.cylinder.length) + '\n')
+            robotFile.write((boundingLevel + 1) * indent + 'height ' + str(boundingObject.geometry.cylinder.height) + '\n')
```
omichel, line 274: Also here we shouldn't write the radius if it is the default value. Same for the height.

urdf2webots/writeRobot.py
```diff
@@ -374,7 +385,7 @@ def URDFVisual(robotFile, visualNode, level, normal=False):
         elif visualNode.geometry.cylinder.radius != 0:
             robotFile.write((shapeLevel + 1) * indent + 'geometry Cylinder {\n')
             robotFile.write((shapeLevel + 2) * indent + 'radius ' + str(visualNode.geometry.cylinder.radius) + '\n')
-            robotFile.write((shapeLevel + 2) * indent + 'height ' + str(visualNode.geometry.cylinder.length) + '\n')
+            robotFile.write((shapeLevel + 2) * indent + 'height ' + str(visualNode.geometry.cylinder.height) + '\n')
```
omichel, line 388: Same here.

omichel (changes requested):
My comments may also apply to other default values, like Sphere.radius, Box.size, etc.

omichel (approved):
Great. Thank you.

PR #160, opened 2022-02-17
Parse clip element from gazebo and add it as near element in proto

astumpf:
Parsing the clip field from the URDF and add it to the lidar node as "near" property. This change makes the lidar respect minimal sense range as defined by the URDF.

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -883,8 +883,11 @@ def parseGazeboElement(element, parentLink, linkList):
                         if lidar.maxRange:
                             lidar.noise /= lidar.maxRange
                 # minRange and near default values are 0.01 in Webots; ensure constraint near <= minRange
-                if not lidar.near and lidar.minRange < 0.01:
+                if lidar.near and lidar.minRange and lidar.near > lidar.minRange:
+                    lidar.minRange = lidar.near
+                    print('The "minRange" value cannot be strictly inferior to the "near" value for a lidar, "minRange" has been set to the value of "near".')
+                elif not lidar.near and lidar.minRange < 0.01:
                     lidar.near = lidar.minRange
-                if not lidar.minRange and lidar.near > 0.01:
+                elif not lidar.minRange and lidar.near > 0.01:
                     lidar.minRange = lidar.near
             Lidar.list.append(lidar)
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -382,10 +382,7 @@ class Lidar():
         if self.numberOfLayers:
             file.write(indentationLevel * indent + '  numberOfLayers %d\n' % self.numberOfLayers)
         if self.minRange:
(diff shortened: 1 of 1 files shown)


urdf2webots/parserURDF.py
```diff
@@ -396,15 +397,18 @@ def export(self, file, indentationLevel):
         if self.numberOfLayers:
             file.write(indentationLevel * indent + '  numberOfLayers %d\n' % self.numberOfLayers)
         if self.minRange:
-            if self.minRange < 0.01:
-                file.write(indentationLevel * indent + '  near %lf\n' % self.minRange)
```
BenjaminHug8, line 400: `minRange` and `near` default values are `0.01` in Webots, with the constraint that `near <= minRange`.

Thus we should set:
-  `near = minRange` if `minRange` < 0.01 and `near` is not defined.
-  `minRange = near` if `near` > 0.01 and `minRange` is not defined.

BenjaminHug8 (changes requested):
Thank you for contribution!

Can you target the `develop` branch with this PR as this is not a fix?

urdf2webots/parserURDF.py
astumpf: Thanks for your feedback! Implemented in https://github.com/cyberbotics/urdf2webots/pull/160/commits/136a429a934a4d1cb6f6db4647f689ae3fd60582

urdf2webots/parserURDF.py
```diff
@@ -877,4 +882,9 @@ def parseGazeboElement(element, parentLink, linkList):
                         lidar.noise = float(noiseElement.getElementsByTagName('stddev')[0].firstChild.nodeValue)
                         if lidar.maxRange:
                             lidar.noise /= lidar.maxRange
+                # minRange and near default values are 0.01 in Webots; ensure constraint near <= minRange
+                if not lidar.near and lidar.minRange < 0.01:
+                    lidar.near = lidar.minRange
+                if not lidar.minRange and lidar.near > 0.01:
+                    lidar.minRange = lidar.near
```
BenjaminHug8, line 889: That is better, we just need to also ensure the constraint when both values are set.

I think it is better to increase `minRange` rather than decrease `near`, as a too low `near` might render parts of the lidar during measurements.

```suggestion
                # minRange and near default values are 0.01 in Webots; ensure constraint near <= minRange
                if lidar.near and lidar.minRange and lidar.near > lidar.minRange:
                    lidar.minRange = lidar.near
                    print('The "minRange" value cannot be strictly inferior to the "near" value for a lidar, "minRange" has been set to the value of "near".')
                elif not lidar.near and lidar.minRange < 0.01:
                    lidar.near = lidar.minRange
                elif not lidar.minRange and lidar.near > 0.01:
                    lidar.minRange = lidar.near
```

BenjaminHug8 (approved):
Thank you!

I let you squash and merge this PR :)

PR #161, opened 2022-02-17
Added the RangeFinder class to represent a depth camera

astumpf:
Add the rangefinder class that reads in a depth camera node from URDF and translates it as a rangefinder node into webots proto.

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -369,6 +369,7 @@ class RangeFinder():
         self.maxRange = None
         self.resolution = None
         self.noise = None
+        self.isImager = True
 
     def export(self, file, indentationLevel):
         """Export this range finder."""
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -353,6 +353,50 @@ class Camera():
         file.write(indentationLevel * indent + '}\n')
 
 
+class RangeFinder():
+    """Define a range finder sensor."""
+
+    list = []
+
+    def __init__(self):
+        """Initialization."""
+        self.name = 'rangefinder'
+        self.fov = None
+        self.width = None
+        self.height = None
+        self.noise = None
+        self.near = None
+        self.minRange = None
+        self.maxRange = None
+        self.resolution = None
+
+    def export(self, file, indentationLevel):
+        """Export this range finder."""
+        indent = '  '
+        file.write(indentationLevel * indent + 'RangeFinder {\n')
+        file.write(indentationLevel * indent + '  name "%s"\n' % self.name)
+        if self.fov:
+            file.write(indentationLevel * indent + '  fieldOfView %lf\n' % self.fov)
+        if self.width:
+            file.write(indentationLevel * indent + '  width %d\n' % self.width)
+        if self.height:
+            file.write(indentationLevel * indent + '  height %d\n' % self.height)
+        if self.noise:
+            file.write(indentationLevel * indent + '  noise %lf\n' % self.noise)
+        if self.near:
+            file.write(indentationLevel * indent + '  near %lf\n' % self.near)
+        if self.minRange:
+            if self.minRange < self.near:
+                file.write(indentationLevel * indent + '  minRange %lf\n' % self.near)
+            else:
+                file.write(indentationLevel * indent + '  minRange %lf\n' % self.minRange)
+        if self.maxRange:
+            file.write(indentationLevel * indent + '  maxRange %lf\n' % self.maxRange)
+        if self.resolution:
+            file.write(indentationLevel * indent + '  resolution %lf\n' % self.resolution)
+        file.write(indentationLevel * indent + '}\n')
+
+
 class Lidar():
     """Define a lidar sensor."""
 
@@ -875,6 +919,37 @@ def parseGazeboElement(element, parentLink, linkList):
                 if hasElement(noiseElement, 'stddev'):
                     camera.noise = float(noiseElement.getElementsByTagName('stddev')[0].firstChild.nodeValue)
             Camera.list.append(camera)
+        elif sensorElement.getAttribute('type') == 'depth':
+            rangefinder = RangeFinder()
+            rangefinder.parentLink = parentLink
+            rangefinder.name = sensorElement.getAttribute('name')
+            if hasElement(sensorElement, 'camera'):
+                cameraElement = sensorElement.getElementsByTagName('camera')[0]
+                if hasElement(cameraElement, 'horizontal_fov'):
+                    rangefinder.fov = float(cameraElement.getElementsByTagName('horizontal_fov')[0].firstChild.nodeValue)
+                if hasElement(cameraElement, 'image'):
+                    imageElement = cameraElement.getElementsByTagName('image')[0]
+                    if hasElement(imageElement, 'width'):
+                        rangefinder.width = int(imageElement.getElementsByTagName('width')[0].firstChild.nodeValue)
+                    if hasElement(imageElement, 'height'):
+                        rangefinder.height = int(imageElement.getElementsByTagName('height')[0].firstChild.nodeValue)
+                if hasElement(cameraElement, 'clip'):
+                    clipElement = cameraElement.getElementsByTagName('clip')[0]
+                    if hasElement(clipElement, 'near'):
+                        rangefinder.near = float(clipElement.getElementsByTagName('near')[0].firstChild.nodeValue)
+            if hasElement(sensorElement, 'noise'):
+                noiseElement = sensorElement.getElementsByTagName('noise')[0]
+                if hasElement(noiseElement, 'stddev'):
+                    rangefinder.noise = float(noiseElement.getElementsByTagName('stddev')[0].firstChild.nodeValue)
+            if hasElement(sensorElement, 'range'):
+                rangeElement = sensorElement.getElementsByTagName('range')[0]
+                if hasElement(rangeElement, 'min'):
+                    rangefinder.minRange = float(rangeElement.getElementsByTagName('min')[0].firstChild.nodeValue)
+                if hasElement(rangeElement, 'max'):
+                    rangefinder.maxRange = float(rangeElement.getElementsByTagName('max')[0].firstChild.nodeValue)
+                if hasElement(rangeElement, 'resolution'):
+                    rangefinder.resolution = float(rangeElement.getElementsByTagName('resolution')[0].firstChild.nodeValue)
+            RangeFinder.list.append(rangefinder)
         elif sensorElement.getAttribute('type') == 'ray' or sensorElement.getAttribute('type') == 'gpu_ray':
             lidar = Lidar()
             lidar.parentLink = parentLink
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -939,6 +939,8 @@ def parseGazeboElement(element, parentLink, linkList):
                 noiseElement = sensorElement.getElementsByTagName('noise')[0]
                 if hasElement(noiseElement, 'stddev'):
                     rangefinder.noise = float(noiseElement.getElementsByTagName('stddev')[0].firstChild.nodeValue)
+                    if rangefinder.maxRange:
+                            rangefinder.noise /= rangefinder.maxRange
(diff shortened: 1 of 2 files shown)


urdf2webots/parserURDF.py
```diff
@@ -949,6 +990,42 @@ def parseGazeboElement(element, parentLink, linkList):
...
+                    if hasElement(imageElement, 'height'):
+                        rangefinder.height = int(imageElement.getElementsByTagName('height')[0].firstChild.nodeValue)
+                if hasElement(cameraElement, 'clip'):
+                    clipElement = cameraElement.getElementsByTagName('clip')[0]
+                    if hasElement(clipElement, 'near'):
+                        rangefinder.near = float(clipElement.getElementsByTagName('near')[0].firstChild.nodeValue)
+            if hasElement(sensorElement, 'noise'):
+                noiseElement = sensorElement.getElementsByTagName('noise')[0]
+                if hasElement(noiseElement, 'stddev'):
+                    rangefinder.noise = float(noiseElement.getElementsByTagName('stddev')[0].firstChild.nodeValue)
```
BenjaminHug8, line 1014:
```suggestion
                    rangefinder.noise = float(noiseElement.getElementsByTagName('stddev')[0].firstChild.nodeValue)
                    if rangefinder.maxRange:
                            rangefinder.noise /= rangefinder.maxRange
```

With this change, the `noise` element should be parsed after the `range` one.

urdf2webots/parserURDF.py
```diff
@@ -949,6 +990,42 @@ def parseGazeboElement(element, parentLink, linkList):
...
+                    rangefinder.minRange = float(rangeElement.getElementsByTagName('min')[0].firstChild.nodeValue)
+                if hasElement(rangeElement, 'max'):
+                    rangefinder.maxRange = float(rangeElement.getElementsByTagName('max')[0].firstChild.nodeValue)
+                if hasElement(rangeElement, 'resolution'):
+                    rangefinder.resolution = float(rangeElement.getElementsByTagName('resolution')[0].firstChild.nodeValue)
+            # minRange and near default values are 0.01 in Webots; ensure constraint near <= minRange
+            if not rangefinder.near and rangefinder.minRange < 0.01:
+                rangefinder.near = rangefinder.minRange
+            if not rangefinder.minRange and rangefinder.near > 0.01:
+                rangefinder.minRange = rangefinder.near
```
BenjaminHug8, line 1027: Same remark as for [here](https://github.com/cyberbotics/urdf2webots/pull/160/files#r811815242)

BenjaminHug8 (changes requested):
Thank you once again for your contribution!

I let you check the requested changes.

BenjaminHug8:
This PR should also target `develop` instead of `master`.

urdf2webots/parserURDF.py
astumpf: Changes applied as commented

astumpf:
Still have some minor issues with the rangefinder, which I hope you can figure out quickly.

First, the resulting proto:
 
```
RangeFinder {
        name "d455_depth_camera"
        fieldOfView 1.518436
        width 1280
        height 720
        near 0.520000
        minRange 0.520000
        maxRange 4.000000
      }
```

Here's the result in RViz:

Screenshot from 2022-02-23 17-32-12

There are some projection issues. The RGBD image in RViz (greyscaled, top left) is identical to the depth image displayed by Webots itself. Therefore, I would argue that this issue happens already in Webots.

astumpf:
> Still have some minor issues with the rangefinder, which I hope you can figure out quickly.
> ...

Solved the transformation issue. The problem was the missing exports of the optical frames in our URDF, so this mistake was on my side. But it requires still another bugfix (see #164).

However, warping remains a serious issue:

Screenshot from 2022-02-23 18-51-22

urdf2webots/parserURDF.py
BenjaminHug8: This part is still missing.

BenjaminHug8:
I will take a look at this, may I ask you your minimal example URDF file so I can also compare with the same example as you have?

urdf2webots/parserURDF.py
astumpf: Yes, didn't had time to review these changes on my side.

BenjaminHug8:
May I ask you your URDF used to generate this PROTO RangeFinder?

astumpf:
> May I ask you your URDF used to generate this PROTO RangeFinder?

I will send you some files the next days.

BenjaminHug8:
After some tests and according with the [documentation on the RangeFinder](https://www.cyberbotics.com/doc/reference/rangefinder), I think the `near` value is too high in your case. This value should be just big enough to exclude any geometry of the RangeFinder from the frustum.

Could you try to reduce it and report here the effect?

astumpf:
> After some tests and according with the [documentation on the RangeFinder](https://www.cyberbotics.com/doc/reference/rangefinder), I think the `near` value is too high in your case. This value should be just big enough to exclude any geometry of the RangeFinder from the frustum.
> 
> Could you try to reduce it and report here the effect?

Set near (and in a second run minRange) to 0.1. Both tests results in following:

Screenshot from 2022-02-23 17-42-33

Curvature is removed, but plane is tilted. I'm just wondering that at max range definite values are reported. According to your [documentation](https://www.cyberbotics.com/doc/reference/rangefinder) +inf should be returned for those pixels, which also matches the [ROS REP117](https://www.ros.org/reps/rep-0117.html). Therefore, we should not be even able to "see" the far plane.

astumpf:
> After some tests and according with the [documentation on the RangeFinder](https://www.cyberbotics.com/doc/reference/rangefinder), I think the `near` value is too high in your case. This value should be just big enough to exclude any geometry of the RangeFinder from the frustum.
> 
> Could you try to reduce it and report here the effect?

Set near (and in a second run minRange) to 0.1. Both tests results in following:

Screenshot from 2022-02-24 15-17-09

Curvature is removed, but plane is tilted. I'm just wondering that at max range definite values are reported. According to your [documentation](https://www.cyberbotics.com/doc/reference/rangefinder) +inf should be returned for those pixels, which also matches the [ROS REP117](https://www.ros.org/reps/rep-0117.html). Therefore, we should not be even able to "see" the far plane.

BenjaminHug8:
> Curvature is removed,

Nice!

> but plane is tilted.

I will investigate that.

> `+inf` should be returned for those pixels

That's something we recently fixed. Are you using Webots R2022a or a nightly build?

astumpf:
> > Curvature is removed,
> 
> Nice!
> 
> > but plane is tilted.
> 
> I will investigate that.
> 
> > `+inf` should be returned for those pixels
> 
> That's something we recently fixed. Are you using Webots R2022a or a nightly build?

Webots R2022a

BenjaminHug8:
> Webots R2022a

Ok then trying one of the [latest nightly build](https://github.com/cyberbotics/webots/releases) should resolve the max range values.

I will get back to you once I have understand the tilt of the plane.

BenjaminHug8:
I made a quick test with the rangeFinder and the result shows a bending for horizontal planes (see underneath). It seems to be affected by the near plane and is probably due to some mapping of the frustum of the camera, I am currently working on it.



The range finder used is the one you gave but with a near plane of 0.1:
```
RangeFinder {
        name "d455_depth_camera"
        fieldOfView 1.518436
        width 1280
        height 720
        near 0.1
        minRange 0.1
        maxRange 4.000000
      }
```

But otherwise I am not able to reproduce the tilt of the plane.

astumpf:
> I made a quick test with the rangeFinder and the result shows a bending for horizontal planes (see underneath, the wall on the right is not the max range but a physical wall). It seems to be affected by the near plane and is probably due to some mapping of the frustum of the camera, I am currently working on it.
> 
> Screenshot from 2022-02-28 11-48-10
> 
> The range finder used is the one you gave but with a near plane of 0.1:
> 
> ```
> RangeFinder {
>         name "d455_depth_camera"
>         fieldOfView 1.518436
>         width 1280
>         height 720
>         near 0.1
>         minRange 0.1
>         maxRange 4.000000
>       }
> ```
> 
> But otherwise I am not able to reproduce the tilt of the plane.

It seems your model has also some tilted coordinate systems. Are you sure you are not rendering the depth image from one of these that could accidently compensate the tilt of the plane?
However, I've send you our proto file a few days ago.

BenjaminHug8:
> It seems your model has also some tilted coordinate systems. Are you sure you are not rendering the depth image from one of these that could accidently compensate the tilt of the plane?

I can check this.

> However, I've send you our proto file a few days ago.

I ll assume that I just have to replace the lidar by a rangefinder in the PROTO file you sent.

astumpf:
> I ll assume that I just have to replace the lidar by a rangefinder in the PROTO file you sent.

The proto is ready to run out of the box :-). No changes required.

BenjaminHug8:
> The proto is ready to run out of the box :-). No changes required.

I am a bit confused, we are talking about rangeFinder in this PR and I received a Lidar.box.proto with a lidar device in it.

Were you talking about a lidar all this time or did I not have the correct file?
(The visual of the PROTO is the same, but I just surprised to found a lidar and not a rangeFinder, I will test with the lidar until you see this message).

astumpf:
> > The proto is ready to run out of the box :-). No changes required.
> 
> I am a bit confused, we are talking about rangeFinder in this PR and I received a Lidar.box.proto with a lidar device in it.
> 
> Were you talking about a lidar all this time or did I not have the correct file? (The visual of the PROTO is the same, but I just surprised to found a lidar and not a rangeFinder, I will test with the lidar until you see this message).

I'm sorry for any confusion. This proto is indeed the "Lidarbox" model I'm using for the tests above. It includes both, Lidar and Rangefinder. Lidar works already fine for us.

BenjaminHug8:
> I'm sorry for any confusion. This proto is indeed the "Lidarbox" model I'm using for the tests above. It includes both, Lidar and Rangefinder. Lidar works already fine for us.

Ah ok, I understand better now. I made a quick test with the rangeFinder and I do not see a tilt of the plane.

However if I rotate the device in Webots it appears that the visualization in RViz is not simply rotated, e.g. if I have a horizontal plane with a vertical wall, after a rotation in of the device in Webots I can see (of course) a rotation in Rviz but the plane and the wall are no longer perpendicular and I almost get a horizontal plane tilted but the wall is still vertical.

So I am wondering if you publish your tf tree from your URDF or if you let the `webots_ros2_driver` do it for you ([here is an example on how to set this option](https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_tiago/launch/robot_launch.py#L77))?
If you use your own URDF I suspect the converter tool to maybe not correctly handle rotations.

BenjaminHug8:
> I made a quick test with the rangeFinder and the result shows a bending for horizontal planes (see underneath, the wall on the right is not the max range but a physical wall). It seems to be affected by the near plane and is probably due to some mapping of the frustum of the camera, I am currently working on it.
> 
> Screenshot from 2022-02-28 11-48-10

This effect has been fixed in https://github.com/cyberbotics/webots/pull/4309.

astumpf:
To give this topic new input. I'm currently experimenting with the latest changes introduced with the 2022b release.

This issue here unfortunately remains. Here is a new example when using a range finder:

```
RangeFinder {
  name "vision_d455_depth_camera"
  fieldOfView 1.518436
  width 1280
  height 720
  near 0.520000
  minRange 0.520000
  maxRange 4.000000
}
```

Webots world:
Screenshot from 2022-10-05 22-00-33

Produced depth image (converted to point cloud using [depth_image_proc](http://wiki.ros.org/depth_image_proc), but the rangefinder image is confirmed to contain the same data):
Screenshot from 2022-10-05 22-00-12

Now, a big concern which is hidden in this example: The color image is perfectly projected into the point cloud. This is an indication that the color image may be heavily warped as well. This would explain the strange behavior that AR marker detectors (e.g. [Aruco](https://docs.opencv.org/4.x/d5/dae/tutorial_aruco_detection.html)) have a hard time to detect anything in camera images rendered by Webots.

astumpf:
Nevermind, this projection issue was on our side due to a wrongly published camera info. After sorting out this issue, everything works perfectly. This PR is ready to merge (finally!) :+1:

omichel (approved):
Looks good to me.
Thank you.

Issue #162, opened 2022-02-22
NameError: global name 'PackageNotFoundError' is not defined


DrakeProvost:
### System information
* Ubuntu 20.04.4 LTS
* Python 2.7.18
* Python 3.8.10

### Installation of `urdf2webots`
```
$ curl https://bootstrap.pypa.io/pip/2.7/get-pip.py | python
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1863k  100 1863k    0     0  9411k      0 --:--:-- --:--:-- --:--:-- 9411k
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
Defaulting to user installation because normal site-packages is not writeable
Collecting pip<21.0
  Downloading pip-20.3.4-py2.py3-none-any.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 11.3 MB/s 
Collecting setuptools<45
  Downloading setuptools-44.1.1-py2.py3-none-any.whl (583 kB)
     |████████████████████████████████| 583 kB 11.0 MB/s 
Collecting wheel
  Downloading wheel-0.37.1-py2.py3-none-any.whl (35 kB)
Installing collected packages: pip, setuptools, wheel
  WARNING: The scripts pip, pip2 and pip2.7 are installed in '/home/drakeprovost/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The scripts easy_install and easy_install-2.7 are installed in '/home/drakeprovost/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script wheel is installed in '/home/drakeprovost/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
ERROR: pip's legacy dependency resolver does not consider dependency conflicts when selecting packages. This behaviour is the source of the following dependency conflicts.
launch-ros 0.11.6 requires pyyaml, which is not installed.
[...]

PR #163, opened 2022-02-23
Update contributing and python command

BenjaminHug8:
Refers to https://github.com/cyberbotics/urdf2webots/issues/162#issuecomment-1048514973

### DIFF
--- a/urdf2webots/importer.py
+++ b/urdf2webots/importer.py
@@ -3,11 +3,16 @@
 """URDF files to Webots PROTO or Robot node converter."""
 
 
+import sys
+
+# Check version of Python
+if sys.version_info < (3,5):
+    sys.exit('This converter tool requires Python 3.5 or higher.')
+
 import errno
 import optparse
 import os
 import re
-import sys
 import tempfile
 from xml.dom import minidom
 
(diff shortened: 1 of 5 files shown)


omichel (commented):
I wouldn't document to call python3 explicitly as this won't work on all systems (such as Windows) where the python3 command doesn't exist. On recent systems calling `python` should invoke python 3 in any case. So instead of documenting to call python 3, I would simply add a test at the beginning of the script to display an error if python < 3 (or even 3.5 or the like) is used and quit. See for example https://stackoverflow.com/a/446136/810268

urdf2webots/importer.py
```diff
@@ -3,11 +3,16 @@
 """URDF files to Webots PROTO or Robot node converter."""
 
 
+import sys
+
+# Check version of Python
+if sys.version_info < (3,5):
+    sys.exit('This converter tool requires Python 3.5 or higher.')
```
omichel, line 10:
```suggestion
    sys.exit('urdf2webots requires Python 3.5 or higher.')
```

omichel (changes requested):
Thank you.
The error message would be clearer if referring to the name of the tool instead.

omichel (approved):
Thank you.

PR #164, opened 2022-02-23
Imaging device requires additional transform as Webots uses different coordinate system for cameras

astumpf:
[Webots](https://cyberbotics.com/doc/reference/camera) defines the camera frame differently as [ROS](http://wiki.ros.org/image_pipeline/CameraInfo). This patch applies the required transform when the sensor is tagged as "imager".

Note: This patch is required for #161

### DIFF
--- a/urdf2webots/writeRobot.py
+++ b/urdf2webots/writeRobot.py
@@ -131,7 +131,16 @@ def URDFLink(robotFile, link, level, parentList,
                 if not haveChild:
                     haveChild = True
                     robotFile.write((level + 1) * indent + 'children [\n')
-                sensor.export(robotFile, level + 2)
+                if hasattr(sensor, 'isImager') and sensor.isImager:
+                    robotFile.write((level + 2) * indent + 'Transform {\n')
+                    robotFile.write((level + 3) * indent + 'translation 0 0 0\n')
+                    robotFile.write((level + 3) * indent + 'rotation 0.577350 -0.577350 0.577350 2.094395\n')
+                    robotFile.write((level + 3) * indent + 'children [\n')
+                    sensor.export(robotFile, level + 4)
+                    robotFile.write((level + 3) * indent + ']\n')
+                    robotFile.write((level + 2) * indent + '}\n')
+                else:
+                    sensor.export(robotFile, level + 2)
         # 3: export Joints
         for joint in jointList:
             if joint.parent == link.name:
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -348,6 +348,7 @@ class Camera():
         self.width = None
         self.height = None
         self.noise = None
+        self.isImager = True
 
     def export(self, file, indentationLevel):
         """Export this camera."""


BenjaminHug8:
I am not so sure that we should change the orientation of the camera, but rather change in `webots_ros2` what we publish in the `CameraInfoMessage` here:

https://github.com/cyberbotics/webots_ros2/blob/ebf6c40e890eb295a8cb15aac3b79b4f61e5f472/webots_ros2_driver/src/plugins/static/Ros2Camera.cpp#L38-L60

@omichel May I ask your opinion on this?

astumpf:
> I am not so sure that we should change the orientation of the camera, but rather change in `webots_ros2` what we publish in the `CameraInfoMessage` here:
> 
> https://github.com/cyberbotics/webots_ros2/blob/ebf6c40e890eb295a8cb15aac3b79b4f61e5f472/webots_ros2_driver/src/plugins/static/Ros2Camera.cpp#L38-L60
> 
> @omichel May I ask your opinion on this?

Note here: If you not change the physical orientation, the image will be rendered along the wrong axes as well. Furthermore, changing the webot_ros2_driver will exclude ROS 1 users (like we).

BenjaminHug8:
> Note here: If you not change the physical orientation, the image will be rendered along the wrong axes as well.

Do you mean the rendering in RViz or in Webots (or both)?

> Furthermore, changing the webot_ros2_driver will exclude ROS 1 users (like we).

That's a good point, I have to check if something might be changed in the ROS 1 interface.

astumpf:
> > Note here: If you not change the physical orientation, the image will be rendered along the wrong axes as well.
> 
> Do you mean the rendering in RViz or in Webots (or both)?

Only in Webots. Removing this patch results in:

Screenshot from 2022-02-24 15-57-02

Screenshot from 2022-02-24 15-57-46

As noted in the initial post, this misalignment matches exactly the transformation between the frame coordinate systems definition of Webots and ROS.

astumpf:
For clarification, here the coordinate systems for the optical sensor frames from which the images in webots should be rendered from:

Screenshot from 2022-02-24 16-12-42

These frames are matching the [ROS conventions](http://wiki.ros.org/image_pipeline/CameraInfo). The URDF is also [provided by Intel](https://github.com/IntelRealSense/realsense-ros/tree/development/realsense2_description) this way.

BenjaminHug8:
Thank you for the information, I will take a look tomorrow!

urdf2webots/writeRobot.py
```diff
@@ -129,7 +129,16 @@ def URDFLink(robotFile, link, level, parentList,
...
                     robotFile.write((level + 1) * indent + 'children [\n')
-                sensor.export(robotFile, level + 2)
+                if hasattr(sensor, 'isImager') and sensor.isImager:
+                    robotFile.write((level + 2) * indent + 'Transform {\n')
+                    robotFile.write((level + 3) * indent + 'translation 0 0 0\n')
+                    robotFile.write((level + 3) * indent + 'rotation 0.577350 -0.577350 0.577350 2.094395\n')
+                    robotFile.write((level + 3) * indent + 'children [\n')
+                    sensor.export(robotFile, level + 4)
+                    robotFile.write((level + 3) * indent + ']\n')
+                    robotFile.write((level + 2) * indent + '}\n')
```
BenjaminHug8, line 139: I agree on the rotation need, but it would be cleaner to combine it with the precedent rotation rather than adding a transform node in another transform node.

https://github.com/cyberbotics/urdf2webots/blob/3122b3164fd38519e00e9fbc2172fbb02751b836/urdf2webots/writeRobot.py#L120-L123

BenjaminHug8 (dismissed):
Thank you, I just have one request (see above).

And as this target is more a fix rather than a feature, it should target the `master` branch (once this PR is merged, I will merge `master` into `develop`).

As a quick note, the orientation of the camera in Webots R2021b was the same as ROS but it has been changed with the FLU convention used for Webots R2022a.

BenjaminHug8:
This change would also request to change the `getUrdf()` function in Webots that I will handle.

urdf2webots/writeRobot.py
astumpf: Yes, I agree. That was something, I've been looked out in the first place but seems to require many changes.
Doing it this way (as extra transform) does not interfere with the existing frames, which may impact how meshes/shapes are visualizes. Imagine very simple models, which do not consider such optical frames.

BenjaminHug8: I think you just have to use the `combineRotations(rotation1, rotation2)` function to combine the rotations and you do it only for the cameras.

I do not see the reason why this would have an impact on the meshes and shapes, the camera is a sensor contained in a link without shapes or meshes and then will be joint to the link containing the shapes and meshes no?

astumpf: > I think you just have to use the `combineRotations(rotation1, rotation2)` function to combine the rotations and you do it only for the cameras.

Ok, sounds good!

> I do not see the reason why this would have an impact on the meshes and shapes, the camera is a sensor contained in a link without shapes or meshes and then will be joint to the link containing the shapes and meshes no?

I fully agree with you that this is the proper modelling approach. My concern was only that it may brake some poor designed URDFs, which use the same link for the mesh and the camera.

BenjaminHug8: > I fully agree with you that this is the proper modelling approach. My concern only was that it may brake some poorly designed URDFs that use the same link for the mesh and the camera.

Well spotted, with this case we prefer to keep the transform only on the device.

I let you change the targeted branch to `master` and then this PR will be fine!

astumpf: Switched target branch to `master`

astumpf:
> I let you change the targeted branch to `master` and then this PR will be fine!

Switched to `master`. I guess we are ready to merge now :-)?

ad-daniel (approved):
Indeed, thank you!

PR #167, opened 2022-03-28
Export position from --init-pos input argument for prismatic joints

astumpf:
This fixes the import of initial positions for prismatic joints.


ad-daniel (approved):
Nice catch, thank you!

Issue #168, opened 2022-03-28
Base link is missing physics element that breaks Gyro node use


astumpf:
Since 2d620813c999c38be73f4fbd5d49fdb031ed8b0a a major change was applied that prevents to attach a physics node to the root element (base). I've also noticed that the new version does generate a lot less links, I guess there is some kind of simplifications regarding static links involved. However, this breaks any current usage with Gyro sensor (parsed from IMU elements from URDF). Now all Gyro nodes are complaining for a missing physics-enabled parent node.

Diffing the previously generated proto file with the new file reveals following change:

![Screenshot from 2022-03-28 19-18-21](https://user-images.githubusercontent.com/7815721/160452430-c7a252ba-a859-4b45-93f8-8a9b92348a00.png)

Readding these lines, fixes the Gyro problem. Therefore, I'm wondering how to fix that issue in the convertor.

ad-daniel:
Can you provide the urdf you used so as to identify which commit actually broke the behavior?

astumpf:
Unfortunately I cannot disclose our internal URDF. But this issue is related to the physics node of the root link of the robot. When the root link (i.e. base_link) is a fixed link and you attach an IMU to it, you will run in this problem.

Basically, it looks like this

```
<?xml version="1.0" encoding="utf-8"?>
<robot xmlns:xacro="http://ros.org/wiki/xacro">  
  <link name="base_link">
    <inertial>
      <mass value="0.001" />
      <inertia ixx="1e-06" ixy="0" ixz="0" iyy="1e-06" iyz="0" izz="1e-06" />
      <origin xyz="0 0 0" rpy="0 0 0" />
    </inertial>
  </link>

  <link name="substructure_link">
    <visual>
      <origin xyz="0 0 0" rpy="3.14159265358979 0 0" />
    </visual>
    <inertial>
      <mass value="19.071313858032227" />
      <inertia ixx="0.63005133573409289" ixy="-0.00076282161535707747" ixz="0.0017282156658180987" iyy="0.37519285561891186" iyz="-0.0038628958352941107" izz="0.63759031368070762" />
      <origin rpy="0 0 0" xyz="0.00979819299726629 0.00408222060576725 0.130251498956635" />
    </inertial>
  </link>

  <joint name="substructure_joint" type="fixed">
    <parent link="base_link" />
    <child link="substructure_link" />
    <origin xyz="0 0 -0.444601164826685" rpy="0 0 0" />
  </joint>

[...]

PR #165, opened 2022-03-10
Fix dummy links position

BenjaminHug8:
Empty links in URDF are removed when the converter tool exports robot strings (otherwise we get `solid` without physics).

In case we have `Link1-joint1-DummyLink-joint2-Link2` in the URDF, the `joint1` and `joint2` has to be merged together in a third joint to get `Link1-joint3-Link2`.

This PR fix the resulting position when two joints are merged in one, add a test on this functionality and enable this functionality when exporting PROTO robots.

PR #170, opened 2022-04-11
Fix loss of physics

ad-daniel:
**Description**
Fixes https://github.com/cyberbotics/urdf2webots/issues/168.

The reason it appears to be working in develop but not in master is because the bug was introduced in this PR https://github.com/cyberbotics/urdf2webots/pull/165 which targeted master.
Basically, it seems that the aim was to collapse structures like `link1 -- j1 -- link2 -- j2 -- link3` down to `link1 -- j3 -- link3` if and only if link2 happened to be a mass-less dummy link. Issue https://github.com/cyberbotics/urdf2webots/issues/168 however arises from the fact that when one of the collapse candidates arises, we force the parent link to become static and therefore if it had physics, we end-up losing this information by making it static.

The proposed change does fix the issue, but I have to admit I don't really understand why we are forcing the parent to be static in the first place. From my understanding, the dummy link is an issue since we can't really have a non-physical node in-between two physically enabled ones, so removing it makes sense but then either:
- the parent (link1) does have physics: it should be kept or we might break stuff (for instance if the same node has sensor children which require the parent to have physics we can't remove it)
- the parent (link1) does not have physics: then the dummy not having it isn't really an issue since we are regardless in a staticbase scenario (although removing it makes the PROTO cleaner)

In short, I'm not sure to understand exactly which scenario does this approach is meant to address.

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -892,8 +892,13 @@ def removeDummyLinksAndStaticBaseFlag(linkList, jointList, toolSlot):
                     jointList.remove(jointList[childJointIndex])
                 else:
                     # Special case for dummy non-root links used to fix the base of the robot
-                    if isRootLink(jointList[parentJointIndex].parent, childList):
-                        staticBase = True
+                    parentLink = jointList[parentJointIndex].parent
+                    if isRootLink(parentLink, childList):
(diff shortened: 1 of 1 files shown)


omichel:
I believe the goal of #165 was simply to make the proto cleaner.

PR #175, opened 2022-06-17
Merge `master` into `develop`

ad-daniel:
Since basically develop and master progressed in parallel in the past months, going forward it might be better to have only the master branch (since that's what both branches of ros2 rely upon anyway).


omichel (approved):
Thank you.

Issue #178, opened 2022-06-27
Converter prunes URDF branches of sensors


astumpf:
Unfortunately, I found out about this problem only recently. But latest version of the converter is pruning important (static/fixed) links from the URDF where sensors are attached. In this specific case, the optical frames of cameras are truncated.

Current behavior:

Camera frame is supposed to be the optical link that is truncated by the converter. I suspect that Webot has a fallback to attach them to the root link as seen in the following screenshot.

![Screenshot from 2022-06-27 18-02-35](https://user-images.githubusercontent.com/7815721/175983912-f3a3c6f0-30b8-4ba5-89b5-b147304a6199.png)

Expected behavior:

Cheated by changing the URDF to set the camera frame to the last non-truncated link.

![Screenshot from 2022-06-27 18-01-06](https://user-images.githubusercontent.com/7815721/175983885-41947fc6-1f61-4565-8d81-65629a94eccf.png)

TheMangalex:
The RemoveDummyLinksAndStaticBaseFlag function doesn't check if a link has sensors, as they haven't been parsed at that point. 
Additionally a preserveFixedJoint or preserveLink tag would be needed in case some frames  are needed in ros applications.

astumpf:
This issue is currently a major blocker when working with Webots. Any hint or ideas to fix it?

astumpf:
Solved with #189

PR #189, opened 2022-10-05
Prevent pruning of links that are referenced by sensors

astumpf:
Fixes issue #178


omichel (approved):
That looks good to me.
Thank you.

PR #190, opened 2022-10-06
Fixup rangefinder noise calcuation

astumpf:
Fixup rangefinder noise calcuation as required max range is loaded at later point


omichel (approved):
Well spotted!
Thank you.

PR #193, opened 2022-11-14
Replace optparse with argparse and target higher python versions for tests

ad-daniel:
As mentioned here https://github.com/cyberbotics/webots_ros2/issues/515

- Replace optparse with argparse (has been deprecated)
- Since python 3.5 and 3.6 have reached their end of life, the test now only favors higher versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
   tests:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


ygoumaz (approved):
Thank you! As discussed we should now:

1. Create the tag.
2. Synchronize the submodule on the `webots_ros2` side
3. Fix the same problem in `webots_ros2` (master branch)
4. Synchronize master and develop on `webots_ros2`.

PR #196, opened 2023-01-16
Fix test sources

ad-daniel:
When running the script on the webots_ros2 side, it detected several issues in the urdf2webots submodule, but these didn't show on this repo. It turns out a simpler version of the test script was being used. Now it uses the same as on the webots side. Some adjustments were therefore necessary.


omichel (approved):
Thank you.

PR #198, opened 2023-02-27
Use relative path in output when relative in urdf

brettle:
When there is a relative path in the urdf, use a relative path in the output
but make the path in the output refer to the file relative to the output
location.

Add a new test case and change an existing test case which was expecting
absolute paths in the output.


ad-daniel (approved):
Nice, thank you!

PR #205, opened 2023-05-31
Update to Pose node

ygoumaz:
https://github.com/cyberbotics/webots/pull/5978 added the Pose node to Webots. The urdf2webots must be adapted to support this new feature.

### DIFF
--- a/tests/expected/Human.proto
+++ b/tests/expected/Human.proto
@@ -1,19 +1,19 @@
 #VRML_SIM R2023b utf8
 # license: Apache License 2.0
 # license url: http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


omichel (changes requested):
Thank you.
It looks good to me, but I would not have renamed Human to HumanR2023b, instead, simply adding HumanR2023a should be sufficient and it should be removed when we release Webots R2023b.

ygoumaz:
Yes, I agree. However, we also have the `HumanR2022a` file, which allows to test the backwards compatibility, as we can specify a target Webots version (and convert files to older formats). So I don't think we will have to remove it. 

I will update the name of the new one.

omichel (approved):
Thank you.

Issue #204, opened 2023-05-29
Missing parsing of gazebo mesh filename 'file://' prefix


IDavGal:
The `file://` prefix is not being parsed out of the mesh filename in the URDF definition.

I'm thinking this can be fixed by modifying the code in [parserURDF.py](https://github.com/cyberbotics/urdf2webots/blob/c4cccd7d9b65dc8fe555839d311d8bdc30fc946d/urdf2webots/parserURDF.py#L595)  in a similar manner to how the `package://` prefix is addressed there, although both checks should be done before the absolute path checking, since filenames with any of these prefixes would be categorized as relative paths and therefore be affected by the `os.path.normpath()` method, collapsing all consecutive backslashes into a single one. The modified code would look like this:

```python
elif hasElement(geometryElement, 'mesh'):
    meshfile = geometryElement.getElementsByTagName('mesh')[0].getAttribute('filename')
    # hack for gazebo mesh database
    if meshfile.count('package'):
        idx0 = meshfile.find('package://')
        meshfile = meshfile[idx0 + len('package://'):]
    if meshfile.count('file'):
        idx0 = meshfile.find('file://')
        meshfile = meshfile[idx0 + len('file://'):]
    if not os.path.isabs(meshfile):
        # Use the path relative to the output file
        meshfile = os.path.normpath(os.path.relpath(os.path.join(path, meshfile), outputDirectory))
```

omichel:
Thank you for this contribution.
Can you please create a new pull request for proposing this change?

PR #206, opened 2023-06-13
Parse `file://` mesh prefix from mesh filepath

IDavGal:
Address changes mentioned in this [issue](https://github.com/cyberbotics/urdf2webots/issues/204).
- Add support for parsing out the `file://` required by rViz from mesh filepaths
- Normalizes absolute paths after parsing out the prefixes, since when present, the filepath would be considered as relative (filepath starts without `/`)

### DIFF
--- a/urdf2webots/parserURDF.py
+++ b/urdf2webots/parserURDF.py
@@ -594,13 +594,16 @@ def getVisual(link, node, path, outputDirectory):
             link.visual.append(visual)
         elif hasElement(geometryElement, 'mesh'):
             meshfile = geometryElement.getElementsByTagName('mesh')[0].getAttribute('filename')
-            if not os.path.isabs(meshfile):
-                # Use the path relative to the output file
-                meshfile = os.path.normpath(os.path.relpath(os.path.join(path, meshfile), outputDirectory))
             # hack for gazebo mesh database
             if meshfile.count('package'):
                 idx0 = meshfile.find('package://')
                 meshfile = meshfile[idx0 + len('package://'):]
+            elif meshfile.count('file'):
+                idx0 = meshfile.find('file://')
+                meshfile = meshfile[idx0 + len('file://'):]
+            if not os.path.isabs(meshfile):
+                # Use the path relative to the output file
+                meshfile = os.path.normpath(os.path.relpath(os.path.join(path, meshfile), outputDirectory))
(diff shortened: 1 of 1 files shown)


IDavGal:
I've encountered this issue when working with an URDF with a workaround for gazebo in ROS 2.

Unrelated, but apparently there is an issue when spawning a robot using gazebo_ros in ROS 2, specifically with mesh filenames using the `package://` prefix. 
There is a discussion regarding this issue [here](https://answers.ros.org/question/373495/ros2-foxy-gazebo-spawn_entity-systempathscc459-file-or-path-does-not-exist/). 
Using `file://$(find )` extends to an expression similar to `file:///path/to/my/file` as you mention.

The URDF I mention can be found [here](https://github.com/Ekumen-OS/Noah-bot-simulation-ros2/blob/foxy/noah_description/urdf/accurate_base_link.urdf.xacro)

omichel:
I see, therefore, I believe we should support the `file:///` protocol which is generated by xacro from the `file://$(find ` syntax.

urdf2webots/parserURDF.py
```diff
@@ -594,13 +594,16 @@ def getVisual(link, node, path, outputDirectory):
...
             # hack for gazebo mesh database
             if meshfile.count('package'):
                 idx0 = meshfile.find('package://')
                 meshfile = meshfile[idx0 + len('package://'):]
+            elif meshfile.count('file'):
+                idx0 = meshfile.find('file://')
+                meshfile = meshfile[idx0 + len('file://'):]
+            if not os.path.isabs(meshfile):
+                # Use the path relative to the output file
+                meshfile = os.path.normpath(os.path.relpath(os.path.join(path, meshfile), outputDirectory))
```
omichel, line 606: That would seem clearer and cleaner to me:
```suggestion
            if meshfile.startswith('package://'):
                idx0 = meshfile.find('package://')
                meshfile = meshfile[idx0 + len('package://'):]  # package relative path
            elif meshfile.startswith('file:///'):
                idx0 = meshfile.find('file:///')
                meshfile = meshfile[idx0 + len('file://'):]  # we keep the last slash for absolute path
            elif not os.path.isabs(meshfile):
                # Use the path relative to the output file
                meshfile = os.path.normpath(os.path.relpath(os.path.join(path, meshfile), outputDirectory))
```

omichel (changes requested):

PR #208, opened 2023-06-29
Prepare version 2023.1.0

ygoumaz:
I could not find any information on the version naming. I think it would be good to match the version numbering of `webots_ros` and `webots_ros2` using the year, webots realease and minor value.


omichel (approved):
Yes, that seems to be a good idea.

PR #213, opened 2024-03-19
Update importer.py

FrogGuaGuaGua:
In Windows system, the url of *.stls in the output proto file contain \ and /, the \ will cause webots software to be unable to read the proto file. therefore, replace the \ with /


omichel (approved):
Thank you.

